PR #3, opened 2017-08-15
New toolchain for creating banners

gbirke:
- Build banners from JS initialization code, JS libraries, templates and CSS with [webpack](https://webpack.js.org/)
- Modernize Banner Markup and CSS
  - Use flexbox CSS instead of tables.
  - Use [BEM naming schema](http://getbem.com/introduction/) for banner components.
- Use Docker image for builds
- Use webpack-dev-server and Docker image for interactive development of banners

Javascript is mostly copied 1:1 from current banners, only minor modifications for changed markup. README details possible next refactoring steps for JS.


JeroenDeDauw:
image

image

;)

desktop/GlobalBannerSettings.js
```diff
@@ -0,0 +1,14 @@
+module.exports = {
```
wiese, line 1: Do we really want to keep this inside the repo (to theoretically "develop on the train"), or - as we go through central notice anyways - lose this and use it from there (as done via prefix in `webpack.common.js`)?

wiese (approved):
forward ever, backward never

desktop/GlobalBannerSettings.js
gbirke: I'd very much like to keep an "offline development" version that works without a proxied Wikipedia. It'd basically work like this:
* have a webpack.local configuration
* have the file `dist/index_local.html`, that includes an inline JavaScript section with the GlobalBannerSettings
* Use `loader.js`to load the banners

Until we create that, let's keep `GlobalBannerSettings.js` and `loader.js`

PR #5, opened 2017-08-18
Compare different animations

gbirke:


wiese:
We talked about it for a bit and agreed that var2 banner probably should not make it into `master`.

wiese:
Hi,

I looked into the animation issue, went to implement an [alternative solution](https://github.com/wmde/fundraising-banners/tree/animation-transform) (prototype, hence no PR) using [transform translate](https://css-tricks.com/almanac/properties/t/transform/), as this is praised to actually leverage GPU (and support [does not seem to be too bad either](http://caniuse.com/#feat=transforms2d)), only to find out this did not change a thing.

**Then I found the reason**. When `displayBanner()` is first invoked, the banner height
is calculated differently then later (in my case 197px). This number is then used to configure the banner animation, while `addSpace()` later does its own calculation and arrives at a different result (in my case 240px). This, of course, leads to different animations.

Both, position absolute, and transform translate, work equally fine when banner and page offsets work with the same number: 9728831

Reason being: The first calculation is done before the `#centralNotice` (incl. our banner) element is moved out of the page content where it is effected by CSS rules stating `font-size: 0.8em;`. If we move it first, and then calculate the height, everything works out again: 868420e

Fix of the actual problem now in #8, now we can look at which way to animate this with CSS we actually want to take...

PR #7, opened 2017-08-23
Move JavaScript libraries into shared folder

gbirke:
Banner functionality that is relatively independent from markup is moved
into a shared folder where it can be used by all banner channels
(mobile, English, wp.de, etc).

Not in this PR: Remove all jQuery selections except on #WMDE_Banner from the 
shared library. Culprits are some deprecated functions in
banner_functions and wlightbox.

This is for https://phabricator.wikimedia.org/T173918


wiese (approved):
The lightbox script to me seems a bit different than the others in that it does not only encapsulate logic, but is a gui element which is not necessarily reusable across all channels (e.g. mobile) - but it _could_ be (e.g. english).

PR #9, opened 2017-08-24
[T174007] dewiki-ba05

KaiNissen:


desktop/css/styles.pcss
```diff
@@ -258,5 +262,14 @@
         flex: 0 0 18px;
     }
 
+    .click-tracking {
+        display: none;
+    }
+
+    .click-tracking__pixel {
```
gbirke, line 269: I was not sure if hidden images are loaded. [At the moment (Aug 2017) they are](https://stackoverflow.com/questions/12158540/does-displaynone-prevent-an-image-from-loading), but maybe we need to look at it again next year.

Test page: https://www.quirksmode.org/css/displayimg.html

desktop/templates/banner_text.hbs
```diff
@@ -1,15 +1,18 @@
...
 	<i>Liebe Leserinnen und Leser:</i>
 </p>
-<p>
+<p class="text__paragraph">
 	Verzeihen Sie die Störung.
 	{{ campaignDaySentence }}
 	Wikipedia wird von Freiwilligen geschrieben und durch Spenden von durchschnittlich 20 € finanziert.
 	Jetzt sind Sie in Deutschland gefragt.
-	<span class="highlight">
+	<span class="text__paragraph text__paragraph--highlight">
```
gbirke, line 9: If used like this, the naming is not very semantic any more, because the highlight is not really a paragraph. I'd go with a simple `text__highlight`. Yes, the highlighted text is kind of "modified", but since modifiers only apply to blocks and elements (as you have done it correctly here), I would go with a new element.

desktop/templates/banner_text.hbs
```diff
@@ -1,15 +1,18 @@
...
 		Wenn alle, die das jetzt lesen, einen kleinen Beitrag leisten, ist unsere Spendenkampagne in einer Stunde vorüber.
 	</span>
 	Schon der Preis einer Tasse Kaffee würde genügen. Millionen von Menschen nutzen Wikipedia, aber nur ein Bruchteil spendet.
 	Wenn Sie Wikipedia nützlich finden, nehmen Sie sich {{ weekdayPrepPhrase }} {{ customDayName }} bitte eine
 	Minute Zeit und geben Wikipedia mit Ihrer Spende etwas zurück. Vielen Dank!
+	<span class="text__paragraph text__paragraph--hidden-placeholder">
```
gbirke, line 15: Again, I'd use a `text_hidden-placeholder` instead.

gbirke (approved):
Thanks for setting up the new banner. My comments can be addressed for the next one, in a followup.

PR #10, opened 2017-08-31
[T173918] add mobile banner

KaiNissen:
This PR adds the mobile banner and some refactoring. I'm not so sure about what I did in aa28874, please have a look.

[phab:T173918](https://phabricator.wikimedia.org/T173918)


shared/global_banner_settings.js
```diff
@@ -5,7 +5,7 @@ module.exports = {
     'donators-base': 0,
     'appr-donations-per-minute': 10,
     'appr-donators-per-minute': 1,
-    'milion-impressions-per-day': 8,
+    'impressions-per-day-in-million': 8,
```
gbirke: We have to be careful when removing/renaming keys in here, because those changes also must be made in the Wiki version. My suggestion for renaming stuff would be to have both side by side and mark one of them as "deprecated". Only remove keys when we are 100% sure that no banners use the old keys.

mobile/banner_ctrl.js
```diff
@@ -0,0 +1,278 @@
...
+const campaignProjection = new CampaignProjection( {
+	campaignStartDate: new Date( GlobalBannerSettings['campaign-start-date'] ),
+	campaignEndDate: new Date( GlobalBannerSettings['campaign-end-date'] ),
+	baseDonationSum: GlobalBannerSettings['donations-collected-base'],
+	donationAmountPerMinute: GlobalBannerSettings['appr-donations-per-minute'],
+	donorsBase: GlobalBannerSettings['donators-base'],
+	donorsPerMinute: GlobalBannerSettings['appr-donators-per-minute']
+} );
+
+const numeral = require( 'numeral' );
```
gbirke: Thanks for using a library instead of using our home-made number formatting 👍 

Two issues:
1) You forgot to add this as a dependency to package.json. 
2) Before we use `numeral`let's investigate other, more lightweight packages (e.g. https://www.npmjs.com/package/format-number or https://www.npmjs.com/package/number-formatter ). `numeral` minified is 11kb and does much more than number formatting: calculations, format parsing, math.

shared/campaign_projection.js
```diff
@@ -0,0 +1,32 @@
+function getSecondsSinceCampaignStart( that ) {
```
gbirke: I'd rather have `campaignStartDate` and `campaignEndDate` as parameters and drop `that`. It's only one more parameter for `calculateProjection`.

shared/campaign_projection.js
```diff
@@ -0,0 +1,32 @@
...
+}
+
+function CampaignProjection( options ) {
+	this.campaignStartDate = options.campaignStartDate || new Date();
+	this.campaignEndDate = options.campaignEndDate || new Date();
+	this.baseDonationSum = options.baseDonationSum || 0;
+	this.donationAmountPerMinute = options.donationAmountPerMinute || 0;
+	this.donorsBase = options.donorsBase || 0;
+	this.donorsPerMinute = options.donorsPerMinute || 0;
+	this.deviation = options.deviation || Math.floor( ( Math.random() ) + 0.5 - 0.2 );
```
gbirke: I was thinking of dropping the "random factor" altogether. We haven't used in in the last 2 years and for people who reload or compare different browsers it might give the appearance that the numbers go "backward", breaking the illusion. It will also make the code much simpler. But let's discuss this.

gbirke (approved):
Great job, I like the refactorings you made!
Kudos for writing tests!

PR #15, opened 2017-09-01
Convert modifiers to elements

gbirke:
Since the previous "modifiers" did not modify a whole block or element
I think it's better to define them as standalone elements.

Also added some docs on the purpose of the `text__hidden-placeholder`
class.

This addresses the concerns from #9


JeroenDeDauw:
+1

wiese:
Rebased.

PR #17, opened 2017-09-01
Fix Size tracking URLs

gbirke:
The path separator was missing for tracking URLs.

### DIFF
--- a/shared/track_size_issues.js
+++ b/shared/track_size_issues.js
@@ -7,7 +7,7 @@ const $ = require( 'jquery' );
  */
 const getVisibleWikipedia = function ( bannerElement ) {
     return $( window ).height() - bannerElement.height();
-}
+};
 
 /**
  * Check if Banner takes too much screen space and track the incident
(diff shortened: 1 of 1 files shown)


shared/track_size_issues.js
```diff
@@ -37,11 +37,11 @@ const trackSizeIssues = function( bannerElement, trackingLink, trackRatio ) {
         window.outerWidth + 'x' + window.outerHeight
     ].join( '--' );
 
-    var trackUrl = trackingLink + resolutions;
+    var trackUrl = trackingLink.replace( /\/$/, '' ) + '/' + resolutions;
```
JeroenDeDauw, line 40: What does this regex do? Remove single slashes?

shared/track_size_issues.js
```diff
@@ -37,11 +37,11 @@ const trackSizeIssues = function( bannerElement, trackingLink, trackRatio ) {
...
     bannerElement.append( $( '<img id="WMDE_Banner-track-sizes" width="1" height="1">' ).attr( 'src', trackUrl ) );
-}
+};
```
JeroenDeDauw, line 42: Not part of this PR: like this there are two "tracking url" variables. Looks like the trackUrl one can be avoided by inlining.

shared/track_size_issues.js
gbirke: Yes, it removes single slashes. Currently, `trackingLink` has none, but this might change in the future so I "normalize" the URL to be able to be able to always add another path element separated with a slash.

PR #19, opened 2017-09-04
Prototype for dewiki desktop banner test 6

KaiNissen:
PR for [phab:174909](https://phabricator.wikimedia.org/T174909)


desktop/templates/banner_text.hbs
```diff
@@ -1,18 +1,17 @@
-<p class="text__headline">
-	<i>Liebe Leserinnen und Leser:</i>
+<p class="text__headline text__headline--italic">
```
gbirke, line 1: That extra class is not necessary IMHO, `text__headline` is semantic enough and can be formatted italic.

desktop/css/styles.pcss
```diff
@@ -51,11 +51,17 @@
...
+    .text__headline--italic {
+        font-style: italic;
+    }
+
+    .text__paragraph {
         margin: .25em 0;
     }
 
-    .text__paragraph--highlight {
+    .text__paragraph .text-highlight {
```
gbirke, line 64: I'm fine with this, but could also be its own element called `.text--highlight`

PR #21, opened 2017-09-26
T173920 Desktop banner on wikipedia.de

wiese:
* proxy request to wikipedia.de at `/wikipedia.de`, e.g. http://localhost:8084/wikipedia.de?bannertest=Banner/B17WMDE_webpack_prototype&devbanner=B17WMDE_07_170913_ctrl
* less CSS assumptions to work without vector skin, too
* dedicated folder for wikipedia.de (even though they don't have a code diff ATM)
* simplified dev entry point, separated banner lists for different projects
* not influencing banner nodes by [wikipedia.de CSS](https://github.com/wmde/Wikipedia.de/pull/29)
* dev banner loader via GS wiki ([> "Development tool"](https://wiki.wikimedia.de/wiki/Web:Banner)), complete w/ [embarrassing history](https://wiki.wikimedia.de/w/index.php?title=Web:Banner/B17WMDE_webpack_prototype&action=history) of writing JS in wikitext-escaped environment

### DIFF
--- a/campaign_info.toml
+++ b/campaign_info.toml
@@ -1,6 +1,7 @@
 # A Campaign with global settings
 [desktop]
 campaign_tracking = "09-ba-171006"
+preview_link = "/wiki/Wikipedia:Hauptseite?banner=B17WMDE_webpack_prototype&devbanner={{banner}}"
 
 # Banners of the campaign, key after "banners" can be anything
 [desktop.banners.ctrl]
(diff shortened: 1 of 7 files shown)


gbirke:
> should wikipedia.de banners get a dedicated folder (even if they don't have a code diff)?

Yes, because they will differ (in CSS or in behavior) from the current banner on .org. But as we refactor and move more and more code into the `shared` folder, this is not a big issue.

> how do the additional wikipedia.de banner influence `campaign_info.toml` and the banner selection list (dev entry point)?

Ideally, the "preview location" (de.wikipedia.org or wikipedia.de) can be specified in `campaign_info.toml` (maybe with de.wikipedia.org as a default) and  `getMediawikiBannerLinks`
in `loader.js` is renamed and refactored to generate different `pathname` and query parameter names,  depending on the "preview location".

wiese:
Rebased.

gbirke (approved):
👍 
I like the simplifications to the initial page and the structure given by the campaign headers.
Iv'e tested all three campaigns and the proxy works well.

PR #23, opened 2017-09-27
darwin deps not needed on (docker) linux npm sys

wiese:
My systems produces a diff in `package-lock.json` installing via `docker-compose run js-build npm install`. Reason is it contains optional dependencies that only get pulled on mac systems.
This PR removes these.

@gbirke Can you please verify that these do not get re-added on your system, when using the docker-based npm?

```
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@1.1.2 (node_modules/fsevents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@1.1.2: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})
```


gbirke:
Nope, when I run `docker-compose run js-build npm` the `fsevents` library gets added. Should I amend the `package-lock.json`?

wiese:
The `package-lock.json` should be identical and remain unchanged no matter what system we install on.

We are not the only ones encountering this. https://github.com/google/web-starter-kit/issues/796#issuecomment-274065368

Surely there is nothing wrong with an optional dependency, what I am looking into now is how to stop the diff in the lock file from happending... https://github.com/npm/npm/issues/17722

wiese:
Running `npm install --no-save` does not change the lock file (shows "SKIPPING OPTIONAL DEPENDENCY" nonetheless) and behaves closer to what one would expect (as opposed to `npm update`) from install. Bit cumbersome but effective.

wiese:
Please update to npm 5.6.0 to avoid problems with optional dependencies in package-lock.json http://blog.npmjs.org/post/167963735925/v560-2017-11-27 /cc @KaiNissen

Little-Little-Monster:
surprise ！
https://lupeier.com/post/linux-npm-install-eacces/

PR #26, opened 2017-10-09
T173920 Skins as objects

wiese:
* implement different skins' behavior in dedicated classes
* implement wpde banner appearance
* fix monobook banner appearance


wiese:
`NOT_A_MEDIAWIKI_SKIN` is gone.

PR #27, opened 2017-10-09
T175686: Improve Campaign day counting and output

gbirke:
This this is for https://phabricator.wikimedia.org/T175686


shared/campaign_day_sentence.js
```diff
@@ -0,0 +1,55 @@
+const SENTENCES = {
```
wiese, line 1: Obviously these are translations for a very specific purpose, but given we do have a global `Translations` list, do we really want to maintain messages (and a separate `trans` function) in all modules? We could inject a translator instead of the language, right?

shared/campaign_day_sentence.js
```diff
@@ -0,0 +1,55 @@
...
+		if ( this.campaignDays.campaignHasEnded() ) {
+			return '';
+		}
+		const daysUntilCampaignEnds = Math.ceil( this.campaignDays.getSecondsUntilCampaignEnds() / 86400 );
+		const daysSinceCampaignStart = Math.ceil( this.campaignDays.getSecondsSinceCampaignStart() / 86400 );
+
+		if ( daysUntilCampaignEnds === 1 ) {
+			return trans( this.language, 'last_day' );
+		} else if ( daysUntilCampaignEnds <= 7 ) {
+			return trans( this.language, 'only_n_days' ).replace( '{{days}}', daysUntilCampaignEnds );
```
wiese, line 46: Just picking up on translations again; variable interpolation should really be left to a dedicated method (think pluralization and whatnot), ideally 3rd party. We _do_ have handlebars already...

wiese (commented):
Maybe (probably) this has nothing to do with the changes in here, but [the desktop banner](http://localhost:8084/wiki/Wikipedia:Hauptseite?banner=B17WMDE_webpack_prototype&devbanner=B17WMDE_10_171018_ctrl) looks rather distorted in my preview... => #41

PR #29, opened 2017-10-10
T176942 Progress bar

wiese:
For https://phabricator.wikimedia.org/T176942


gbirke:
Rebased to current master. Will need some same changes when #27 is merged.

campaign_info.toml
```diff
@@ -11,7 +11,7 @@ tracking = "org-10-171026-ctrl"
 
 [desktop.banners.var]
 filename = "./desktop/banner_var.js"
-pagename = "B17WMDE_10_171018_var"
+pagename = "B17WMDE_progress_bar_test"
```
wiese: This (as well as the `var` name) probably should not make it into the main line.

desktop/css/styles.pcss
```diff
@@ -299,4 +299,9 @@
 		border: 0 none;
 	}
 
+	/* Progress bar styling */
```
wiese: What is the rational (same for `var` and `mobile`) to do this in desktop CSS instead of relying on progress bar CSS?

mobile/banner_var.js
```diff
@@ -53,7 +56,8 @@ $bannerContainer.html( bannerTemplate( {
 	daysRemaining: BannerFunctions.getDaysRemaining( LANGUAGE ),
 	amountBannerImpressionsInMillion: GlobalBannerSettings[ 'impressions-per-day-in-million' ],
 	CampaignName: CampaignName,
-	BannerName: BannerName
+	BannerName: BannerName,
+	progressBar: progressBar.render()
```
wiese: This was rapid prototyping, to me somehow feels dodgy. Are you happy with it this way? I am thinking about a handlebars helper to make this truly a widget that binds to banner life-cycle, not the other way around (having to know the internals of progress bar).

wiese (commented):
Maybe this never happens in reality, but the demo linked in phab animates the bar to its min width while showing "0,0 mio" which is sort of counter-intuitive.

desktop/css/styles.pcss
gbirke: `styles.pcss`has this more by accident, it should only be in `styles_var.pcss`, as the setup at the moment is that only the `var` banners have the progress bar. When the campaign starts, the progress bar is **not** an A/B test but a feature of all banners.

Ideally, the progress bar CSS should only contain behavior (`display:flex`, `visibility`) and the styling (color, border, border radius, margin, padding, font size, line height) should be done inside the styles of the banners. I'll move the styles in #43

gbirke:
Styles have been approved by Till, progress bar can now be used if needed

PR #30, opened 2017-10-12
Hide banners on small viewports

KaiNissen:
This PR is for [phab:T176226](https://phabricator.wikimedia.org/T176226).


wiese (commented):
A good direction - yeay for tests.

I would love to see _one_ place where the following is configured and reused from:
- trackingBaseUrl
- sizeIssueThreshold
- sizeIssueTrackRatio

`trackingLinkGenerator` should better be called `trackingEvents` everywhere, like in the tests.

If `TrackingEvents` contains a dedicated method to track size issues, then the `SizeIssueIndicator` should not be occupied with formatting the piwik request (`x`s and `--`) but just provide the plain data.

As discussed, `SizeIssueIndicator.generateTrackingData` should only receive the banner height, not the entire object (desktop var & ctrl).

Why not create native ES modules, like campaign_days.js in #27 ?

package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Building WMDE banners with webpack",
   "main": "index.js",
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1",
+    "test": "./node_modules/mocha/bin/mocha",
```
wiese: We just need the bin here. @gbirke also [added some babel foo](https://github.com/wmde/fundraising-banners/pull/27/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R7) in his recent feature branch, which I am not sure what it does, but looks clever.

KaiNissen:
I addressed your comments, please have another look.

wiese (approved):

PR #31, opened 2017-10-16
Updated lock file

JeroenDeDauw:
I ran the first 3 commands from the README file and ended up with this modified lock file.


wiese (changes requested):
This is a mac vs linux problem within NPM - formerly discussed in #23.

Until https://github.com/npm/npm/issues/17722 is fixed, please install using `npm install --no-save`

PR #32, opened 2017-10-16
Updated mobile CTRL and VAR for Sofort test

JeroenDeDauw:
For https://phabricator.wikimedia.org/T176720

* Updated CTRL to be VAR
* Renamed Überweisung to Überweisungszusage
* Added Sofort-Überweisung button that goes to fun app with payment type SUB
* SMS button is now 50% width instead of 100%
* CTRL adds pmt=0 to url when sending user to fun app


wiese (commented):
The file moves seem to have completely escaped git(hub) - hard to follow, but the preview renders nicely.

We have to rename campaign and banners (`campaign_info.toml`) before using them in central notice. Let's talk to @KaiNissen about that.

Please remove the `package-lock.json` change from the PR (see #31).

PR #37, opened 2017-10-17
T176720 Sofort mobile

wiese:
* Updated CTRL to be VAR
* Renamed Überweisung to Überweisungszusage
* Added Sofort-Überweisung button that goes to fun app with payment type SUB
* SMS button is now 50% width instead of 100%
* CTRL adds pmt=0 to url when sending user to fun app
* incorporate linting
* new break point for multi-column payment type buttons to address changed label length

PR #38, opened 2017-10-17
dewiki-test-10

KaiNissen:
This is for [phab:T177600](https://phabricator.wikimedia.org/T177600).

### DIFF
--- a/desktop/banner_var.js
+++ b/desktop/banner_var.js
@@ -104,22 +104,11 @@ function validateAndSetPeriod() {
 }
 
 $( '#WMDE_Banner-payment button' ).click( function ( e ) {
-	const paymentType = $( this ).data( 'payment-type' );
-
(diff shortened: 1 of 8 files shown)


JeroenDeDauw:
+1

desktop/banner_var.js
```diff
@@ -1,9 +1,9 @@
-require( './css/styles.pcss' );
+// require( './css/styles.pcss' );
```
wiese, line 1: Not a hard objection, but for mobile (#37) I went the way of _not_ duplicating the ctrl CSS but create a file for var that contains only the delta and load both.

wiese (changes requested):
The var behavior is fine, I believe.

Conceptual concern: doesn't the ctrl require the `pmt=0` GET-parameter?

wiese (approved):

PR #39, opened 2017-10-25
update tracking vars

KaiNissen:
Set tracking vars to actual test start date.


KaiNissen:
Banner code on already updated on meta-wiki.

PR #41, opened 2017-10-27
Fix left padding in infobox

gbirke:
Use rem instead of em to make the padding independent from font size.
This aligns the main text message and the line above the footer.


wiese:
Compensates for #38.

wiese:
As `styles_var.pcss` is not a λ-CSS this only fixes CTRL :(
Took the liberty to add.

wiese (approved):
Thanks a lot!
https://zellwk.com/blog/rem-vs-em/

PR #42, opened 2017-10-27
fix selectors for click tracking events

KaiNissen:
The selectors are referring to elements that don't exist in the DOM.


KaiNissen:
Code has already been deployed to meta.

PR #43, opened 2017-10-27
Refactor progress bar

gbirke:
Rely more on CSS, less on JavaScript.
Use CSS flexbox for distribution of elements.
Use `progress_bar--lateprogress` for when we have many donations. This
will display the text for remaining donations outside the donation bar.

### DIFF
--- a/mobile/css/styles_var.pcss
+++ b/mobile/css/styles_var.pcss
@@ -23,6 +23,18 @@
 		margin: 20px 20px;
 	}
 
+	.progress_bar__donation_remaining--outer {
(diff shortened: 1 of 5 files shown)


gbirke:
TODO: Test in IE 9

JeroenDeDauw:
+0.00000001? :)

gbirke:
This is now ready to be reviewed. 

I'd leave the "global" CSS flag `progress_bar--lateprogress` as it is for now - it does not make the overall code worse (previous version used a global JavaScript property) and is on the list of possible refactorings.

PR #44, opened 2017-11-03
dewiki-test-11

gbirke:
Tests banner hiding on size issues


desktop/css/styles.pcss
```diff
@@ -11,11 +11,10 @@
 	 * CUSTOM
 	 */
 
+	display: none;
 	width: 100%;
 	position: fixed;
 	z-index: 1000;
-	display: none;
-	font-size: 16px;
```
wiese: `font-size: 16px;` was introduced to create a level playing field on wp.de. Why was it removed here?

gbirke: Simple oversight while using an older banner version. Fixed now.

PR #45, opened 2017-11-06
T173919 Implement english banner

gbirke:


desktop/templates/banner_text_var.hbs
```diff
@@ -1,3 +1,4 @@
+<<<<<<< HEAD
```
wiese: Is that good?

desktop/banner_ctrl.js
```diff
@@ -118,7 +118,7 @@ function setupAmountEventHandling() {
 function validateAndSetPeriod() {
 	var selectedInterval = $( '#WMDE_Banner-frequency input[type=radio]:checked' ).val();
 	if ( typeof selectedInterval === 'undefined' ) {
-		BannerFunctions.showFrequencyError( 'Bitte wählen Sie zuerst ein Zahlungsintervall.' );
+		$( '#WMDE_Banner' ).trigger( 'validation:period:error', Translations[ 'no-interval-message' ] );
```
wiese: Here we change both, the translation (understandable) as well as the mechanism to deploy it - switching from dedicated function to pure `$` feels like a deterioration. Is this intentional?

wiese (changes requested):
Mind the unresolved merge conflict!

desktop/banner_ctrl.js
KaiNissen: Hm, I'm not sure whether it is better to hide `$` in the respective function. Might be a good idea for having the event handling in `BannerFunctions` only, though.

In my opinion, what happens in `BannerFunctions` should be dropped/extracted into modules over time anyway. In this case we should add an `EventManager` that cares for handling all events.

wiese (approved):

PR #46, opened 2017-11-09
dewp test12

KaiNissen:
This is for [phab:T180119](https://phabricator.wikimedia.org/T180119).

The code has already been deployed to CentralNotice.


gbirke (commented):
This test is about testing two banner designs, with bold and not-so-bold fonts and subtle form design differences, right?

KaiNissen:
The main difference is, that we use the same font-size and line-height as
in the main content of pages.

gbirke (approved):

PR #51, opened 2017-11-14
Test 01 for wikipedia.de

KaiNissen:
see [phab:T180119](https://phabricator.wikimedia.org/T180119)

Marking this as WIP, since the wrapping mechanism is not proper anymore
- ~~`<nowiki>` tag needs to be removed~~ (fixed by #60)
- ~~banner values for functionality like projection needs to be embedded differently~~ (see [phab:T184283](https://phabricator.wikimedia.org/T184283))

I switched the banner fetching mode on wikipedia.de to `raw` to prevent the script from being made unexecutable by the fetching mode `render`. I also created another [proxy banner](https://wiki.wikimedia.de/wiki/Web:Banner/B17WMDE_webpack_prototype_for_raw_mode) that can be used in `raw` fetch mode.

see also: wmde/Wikipedia.de#31


webpack/wikitext_templates/wikipedia_de.hbs
```diff
@@ -0,0 +1,9 @@
+<!-- WMDE compiled banner, see https://github.com/wmde/fundraising-banners -->
```
wiese, line 1: Having a dedicated wrapper file may be the quickest way to overcome some current problems but it also introduces an increased chance for future problems (change of the data attributes, temptation to add an additional class [for CSS] to only one of them).
That would be a great place to [use a decorator](https://www.sitepoint.com/using-inline-partials-and-decorators-with-handlebars-4-0/#using-decorators) in my eyes.

PR #52, opened 2017-11-16
Disable banners during search

gbirke:


gbirke:
This is a proof of concept for preventing unwanted interaction between the search functionality and banners. If this is approved, apply the changes to CTRL to the English banner and the VAR banners as well before merging.

Ignore the 4 first commits, they were an accidental rebase.

PR #61, opened 2017-11-22
dewiki test20

WMDE-Fisch:
based on branch dewiki-test18

https://phabricator.wikimedia.org/T180950

### DIFF
--- a/desktop/banner_ctrl.js
+++ b/desktop/banner_ctrl.js
@@ -5,8 +5,8 @@ require( './css/wlightbox.css' );
 // BEGIN Banner-Specific configuration
 const bannerCloseTrackRatio = 0.01;
 const sizeIssueThreshold = 180;
-const sizeIssueTrackRatio = 1;
-const searchBoxTrackRatio = 1;
+const sizeIssueTrackRatio = 0.01;
+const searchBoxTrackRatio = 0.01;
 const LANGUAGE = 'de';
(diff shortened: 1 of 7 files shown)


desktop/banner_ctrl.js
```diff
@@ -5,8 +5,8 @@ require( './css/wlightbox.css' );
 // BEGIN Banner-Specific configuration
 const bannerCloseTrackRatio = 0.01;
 const sizeIssueThreshold = 180;
-const sizeIssueTrackRatio = 1;
-const searchBoxTrackRatio = 1;
+const sizeIssueTrackRatio = 0.01;
+const searchBoxTrackRatio = 0.01;
```
wiese, line 9: `banner_ctrl.js` seems not to be using `searchBoxTrackRatio`.

desktop/banner_var.js
```diff
@@ -252,6 +276,8 @@ $( '#ca-ve-edit, .mw-editsection-visualeditor' ).click( function () {
 $( function () {
 	var $bannerElement = $( '#WMDE_Banner' );
 
+	$( 'body' ).prepend( $( '#centralNotice' ) );
```
wiese: Is this a/the fix for the height calculation (and resulting hiding) problem that recently came up again? /cc @gbirke

KaiNissen: @wiese Yes. We (@gbirke and me) tried to think of a better solution, but agreed on putting it here.

PR #62, opened 2017-11-23
initial commit for dewiki-test21

WMDE-Fisch:
based on dewiki-test20 that's why PR is against that branch

https://phabricator.wikimedia.org/T180444


KaiNissen (approved):

JeroenDeDauw:
+1 to @KaiNissen his changes

desktop/templates/banner_html_var.hbs
```diff
@@ -133,6 +133,14 @@
 			</form>
 		</div>
 
+    <div class="form redirect-spinner-overlay" style="display: none">
```
wiese, line 136: `<div class="form` That does not sound right given the very un-formy div-content. Can't find new matching CSS or JS selectors either.
Indentation seems to be spaces instead of tabs (shame CI did not catch this).

PR #68, opened 2017-11-27
[WIP] dewiki test 23

KaiNissen:
see [phab:T181398](https://phabricator.wikimedia.org/T181398)


KaiNissen:
tagged as "work in progress" since the control banner might change depending on the outcome of the previous test

PR #81, opened 2017-12-05
dewiki-xmastree

WMDE-Fisch:
based on dewiki-test22

### DIFF
--- a/desktop/banner_var.js
+++ b/desktop/banner_var.js
@@ -20,10 +20,10 @@ import CampaignDays, { startOfDay, endOfDay } from '../shared/campaign_days';
 import CampaignDaySentence from '../shared/campaign_day_sentence';
 import DayName from '../shared/day_name';
 import InterruptibleTimeout from '../shared/interruptible_timeout';
+import Translations from '../shared/messages/de';
 
 const DevGlobalBannerSettings = require( '../shared/global_banner_settings' );
 const GlobalBannerSettings = window.GlobalBannerSettings || DevGlobalBannerSettings;
-import Translations from '../shared/messages/de';
 const BannerFunctions = require( '../shared/banner_functions' )( GlobalBannerSettings, Translations );
 const campaignDays = new CampaignDays(
 	startOfDay( GlobalBannerSettings[ 'campaign-start-date' ] ),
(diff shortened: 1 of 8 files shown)


desktop/banner_var.js
```diff
@@ -1,9 +1,9 @@
-require( './css/styles.pcss' );
+// require( './css/styles.pcss' );
```
wiese, line 1: This caught my eye, then I thought "sure" - it's a reimplementation so - no delta-CSS, then I looked at the CSS file, now I'm torn again. Maybe we should move the first parts of `styles.pcss` into a dedicated file (`reset.pcss`?)? But that's not a problem of this PR.

desktop/banner_var.js
```diff
@@ -20,16 +20,16 @@ import CampaignDays, { startOfDay, endOfDay } from '../shared/campaign_days';
...
 const DevGlobalBannerSettings = require( '../shared/global_banner_settings' );
 const GlobalBannerSettings = window.GlobalBannerSettings || DevGlobalBannerSettings;
-import Translations from '../shared/messages/de';
 const BannerFunctions = require( '../shared/banner_functions' )( GlobalBannerSettings, Translations );
 const campaignDays = new CampaignDays(
 	startOfDay( GlobalBannerSettings[ 'campaign-start-date' ] ),
 	endOfDay( GlobalBannerSettings[ 'campaign-end-date' ] )
 );
-const campaignDaySentence = new CampaignDaySentence( campaignDays, LANGUAGE, 22 );
+const campaignDaySentence = new CampaignDaySentence( campaignDays, LANGUAGE );
```
wiese, line 32: urgencyThreshold (optional 3rd parameter) was newly introduced. Is it irrelevant because the tree does not render the amount of days left?

wiese (commented):
This works nicely and the responsiveness is great given the amount of elements. You even took care of the custom amount vs package +/- dilemma!

Two minor things:
- on my ubuntu the default sans-serif font is Bitstream Vera Sans Bold which is slightly bigger than average causing the interval labels to run out of the buttons, and the submit button label to wrap
- (delighter) the colored hover packages could be preloaded to make it feel more snappy

KaiNissen:
superseded by and rebased in #96

desktop/banner_var.js
KaiNissen: `urgencyThreshold` defaults to 7. Actually `campaignDaySentence` is not needed at all here.

PR #102, opened 2017-12-27
T183280 Thank you1

wiese:
Not sure if this should be rebased at all or stay a dedicated branch/tag - will likely not profit from the new master, and probably never make into master itself. 

ecadd2c brought a blunt over-exaggeration of `appr-donations-per-minute` to force a 100% progress bar.

Depending on your thoughts, ca5765b and 50ac15e could be later cherry-picked into the main line for future use.


gbirke:
If you rebase this, please keep the banner as VAR, I'm using CTRL in #104 and the two banners will be in the same campaign, but have completely different markup and style.

wiese:
Not sure if this should be rebased at all or stay a dedicated branch/tag - will likely not profit from the new master, and probably never make into master itself. 

ecadd2c brought a blunt over-exaggeration of `appr-donations-per-minute` to force a 100% progress bar.

Depending on your thoughts, ca5765b and 50ac15e could be later cherry-picked into the main line for future use.

gbirke (approved):
I tried several breakpoints and the contents looked good on each one. ACs are fulfilled.

KaiNissen (approved):

KaiNissen:
Rebased and added some last minute (and even past last minute) changes. I agree to @wiese that this shouldn't be merged into master, but kept for future reference.

wiese:
Tagged as https://github.com/wmde/fundraising-banners/releases/tag/thankYou1

PR #115, opened 2018-04-24
Create Mobile banner with sliding functionality

aniansson:


tzhelyazkova (changes requested):
I believe npm-debug.log shouldn't be part of the changes.
Is package-lock.json needed? I see two new dependencies are added to package.json, which means upon npm install those will be cleanly installed and not updated. Therefor, it seems package-lock.json shouldn't be part of the PR, or at least not with all of the 178 new lines of code.

prototype/banner_ctrl.js
```diff
@@ -0,0 +1,195 @@
+/* eslint no-alert: 1 */
+
+require( './css/styles.pcss' );
```
tzhelyazkova: Please put all the require statements at the top followed by the import statements and then all other declarations and logic.

tzhelyazkova (approved):

campaign_info.toml
```diff
@@ -1,9 +1,13 @@
+# NOTE: The "pagename" value must start with an uppercase "B" and should be followed by the year (YY) and WMDE.
+# Example: "B18WMDE_"
+
```
gbirke: Maybe link to https://github.com/wmde/fundraising-infrastructure/wiki/How-to-manage-banners-on-de.wikipedia.org

gbirke (approved):

PR #117, opened 2018-05-03
Update dependencies to resolve security issues

ghost:
Created another update to resolve security issues with the current set of dependencies. Manually tested that all functionalities were preserved, could not find any issues but a second look would certainly not hurt.


ghost:
Not entirely happy about the second commit but since some of the old code uses two selector IDs in various places I had to configure the max ID selector rule which was enabled by the dependency update to allow 2 ID selectors. This should be refactored to 1 (or better: 0) in the future but for now I consider this a necessary evil to get the dependencies up-to-date.

README.md
```diff
@@ -75,6 +75,7 @@ The changes to the code depend on which kind of test you are running.
 - [ ] Create upload plugin for webpack uses the campaign information from `campaign_info.toml` to upload the generated `.wikitext` file to the appropriate page on meta.wikimedia.org / GS-Wiki.
 
 ## Notes on possible Banner code improvements
+* Refactor old CSS to remove ID selectors and ensure that no new CSS uses ID selectors by setting "selector-max-id" in .stylelintrc.json to 0
```
tzhelyazkova: This description doesn't really match the changes in .stylelintrc.json
Probably you meant "to 2"?

ghost: Hm, maybe I was not clear enough then, I have updated the description. What I meant was that the "selector-max-id" value should be set to 0 after the CSS is refactored to have no more ID selectors. It is currently set it to 2 was because Travis was failing due to linting errors with the old CSS. Previous developers used two-chained ID selectors (like `#elementOne #elementTwo { color: blah; }`) which is unnecessary and is something we should get rid of in the future.

tzhelyazkova: Ah ok then.

tzhelyazkova (approved):

PR #118, opened 2018-06-11
Update Mocha to resolve CVE in sub-dependency

ghost:


ghost:
All tests are still passing on my local, so it should be good.

gbirke:
Thanks for the quick response. 

For the people who were concerned about the security of our banners: This was a dev dependency, the banners themselves were never affected.

PR #119, opened 2018-06-25
Replace amount options by a single text input

tzhelyazkova:
Feature: T197952

Banner Test Desktop 01
Formatting css styles

### DIFF
--- a/campaign_info.toml
+++ b/campaign_info.toml
@@ -4,7 +4,7 @@
 
 # A Campaign with global settings
 [desktop]
-campaign_tracking = "41-ba-171229"
+campaign_tracking = "01-ba-180709"
 preview_link = "/wiki/Wikipedia:Hauptseite?banner=B17WMDE_webpack_prototype&devbanner={{banner}}"
 wrapper_template = "wikipedia_org"
 
@@ -12,13 +12,13 @@ wrapper_template = "wikipedia_org"
 # Banners of the campaign, key after "banners" can be anything
 [desktop.banners.ctrl]
 filename = "./desktop/banner_ctrl.js"
-pagename = "B17WMDE_41_171230_ctrl"
-tracking = "org-41-171229-ctrl"
+pagename = "B18_WMDE_01_ctrl"
+tracking = "org-01-180709-ctrl"
 
 [desktop.banners.var]
 filename = "./desktop/banner_var.js"
-pagename = "B17WMDE_41_171230_var"
-tracking = "org-41-171229-var"
+pagename = "B18_WMDE_01_var"
+tracking = "org-01-180709-var"
 
 [dewiki-default]
 campaign_tracking = "ca17-default"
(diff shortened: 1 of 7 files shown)


campaign_info.toml
```diff
@@ -12,13 +12,13 @@ wrapper_template = "wikipedia_org"
 # Banners of the campaign, key after "banners" can be anything
 [desktop.banners.ctrl]
 filename = "./desktop/banner_ctrl.js"
-pagename = "B17WMDE_41_171230_ctrl"
-tracking = "org-41-171229-ctrl"
+pagename = "B17WMDE_01_180709_ctrl"
```
gbirke: The pagename should match the page name on meta, i.e. `B18_WMDE_01_ctrl`. Apparently we're ditching the date from the names (which makes sense).

Please also change the `campaign_tracking` to `01-ba-180709`.

desktop/css/styles.pcss
```diff
@@ -508,6 +507,10 @@
 			.select-group__custom-input {
 				border-radius: 0 0 0.4em 0;
 			}
+			.select-group__custom-input:focus,
+			.select-group__custom-input:active {
```
ghost: You should move the `:focus` and `:active` blocks into the `.select-group__custom-input` block which is defined right above this one to reduce redundancy:

```
.select-group__custom-input {
    border-radius: 0 0 0.4em 0;
    &:focus, 
    &:active {
        //CSS here
}
```

desktop/css/styles_var.pcss
```diff
@@ -84,3 +83,20 @@
 	}
 
 }
+
+#WMDE_Banner-form {
+	.select-group__custom-input,
+	.select-group__custom-input:focus,
+	.select-group__custom-input:active {
```
ghost: It should be possible to do this without re-writing all rules for all three of those cases and using `important`, especially when working with a single ID selector. 

If there is an issue with getting rules applied to elements of this specific selector, you may want to change `styles.pcss` a bit so that it is easier to overwrite these values. I think the old CSS was not fully optimized in this regard so this may be a good opportunity to change that.

desktop/css/styles_var.pcss
```diff
@@ -84,3 +83,20 @@
...
+		border: 0 !important;
+		border-radius: 0.4em !important;
+		border-bottom: 1px solid #808080 !important;
+		background-color: unset;
+		width: 70%;
+		margin: 0 3em;
+	}
+
+	.euro-sign {
+		right: 1.8em !important;
```
ghost: Same as above, I think this should be possible without the use of `!important`.

desktop/banner_var.js
```diff
@@ -1,10 +1,8 @@
 require( './css/styles.pcss' );
+require( './css/styles_var.pcss' );
 require( './css/icons.css' );
 require( './css/wlightbox.css' );
 
-// For A/B testing different styles, load
-require( './css/styles_var.pcss' );
```
gbirke, line 6: In the past, we always kept this line here, together with the comment, either active or commented out, as a hint on what's happening. But that could also be seen as a bad practice. Let's discuss.

tzhelyazkova: Let's remove this comment and when `styles_var.pcss` is deleted, then this require statement would also be deleted.

PR #120, opened 2018-06-28
Add improvements to c18 banners

tzhelyazkova:
Feature: [T197952](https://phabricator.wikimedia.org/T197952), [T197963](https://phabricator.wikimedia.org/T197963)

In CTRL banner:
- Change placeholder text to `Wunschbetrag`
- Add euro sign after the entered amount

In VAR banner:
- Change custom amount color to dark green on value entered
- The custom amount field appears below the interval buttons.

### DIFF
--- a/desktop/banner_ctrl.js
+++ b/desktop/banner_ctrl.js
@@ -123,7 +123,13 @@ function setupValidationEventHandling() {
 		addSpaceInstantly();
 	} );
 }
-
+function appendEuroSign( field ) {
+	if ( $( field ).val() !== '' &&
+		!/^.*(€)$/.test( $( field ).val() )
+	) {
+		$( field ).val( $( field ).val() + ' €' );
+	}
+}
 function setupAmountEventHandling() {
 	var banner = $( '#WMDE_Banner' );
 	// using delegated events with empty selector to be markup-independent and still have corrent value for event.target
@@ -137,6 +143,10 @@ function setupAmountEventHandling() {
 		$( '#WMDE_Banner-amounts .select-group__input' ).prop( 'checked', false );
 		$( '.select-group__custom-input' ).addClass( 'select-group__custom-input--value-entered' );
 		BannerFunctions.hideAmountError();
+		$( '.select-group__custom-input' ).on( 'blur', function () {
+			var self = this;
+			appendEuroSign( self );
+		} );
 	} );
 
 	banner.on( 'paymenttype:selected', null, function () {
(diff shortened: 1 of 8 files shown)


desktop/banner_ctrl.js
```diff
@@ -125,13 +125,23 @@ function setupValidationEventHandling() {
...
 function setupAmountEventHandling() {
 	var banner = $( '#WMDE_Banner' );
 	// using delegated events with empty selector to be markup-independent and still have corrent value for event.target
 	banner.on( 'amount:selected', null, function () {
 		$( '#amount-other-input' ).val( '' );
 		$( '.select-group__custom-input' ).removeClass( 'select-group__custom-input--value-entered' );
+		$( '.select-group__custom-input' ).on( 'blur', function () {
+			var self = this;
+			appendEuroSign( self );
+		} );
```
gbirke: What was your reasoning behind putting the creation of the new event handler into the event handler when a predefined amount is selected?

I think the call to `appendEuroSign` is better suited be put in the `amount:custom` event handler.

desktop/banner_ctrl.js
```diff
@@ -125,13 +125,23 @@ function setupValidationEventHandling() {
 		addSpaceInstantly();
 	} );
 }
-
+function appendEuroSign( field ) {
+	if ( $( field ).val() !== '' &&
+		!/^.*(€)$/.test( $( field ).val() )
+	) {
+		$( field ).val( $( field ).val() + ' €' );
```
gbirke: Unfortunately, adding the sign to the value of the input field breaks the in-banner validation. It might also create invalid input for the FundraisingFrontend, when the field value is sent to it.

Since input fields are not container (and hence we can't use `:after` CSS rules), the euro sign requirement becomes more difficult. I see three options going forward:
* Remove euro sign on form submission (before validation). If validation does not pass, add it back.
* Add an euro-sign overlay that is positioned according to the content of the input field. Don't know if it's possible to center that properly.
* Add an Euro-Sign overlay/sister element that is positioned right to the input field.

gbirke (changes requested):

desktop/banner_ctrl.js
tzhelyazkova: That was a mistake, it should definitely be in `amount:custom` event handler.

gbirke (approved):

ghost:
`tracking_events.js` is failing purposefully since the click events are not working right now. Undo changes marked by `TODO` before merging.

PR #122, opened 2018-07-11
Add ADR for version control strategy

gbirke:
Add document for architectural design decision
(http://thinkrelevance.com/blog/2011/11/15/documenting-architecture-decisions)
outlining how we branch, commit and tag in this repository.

### DIFF
--- a/doc/adr/001-version-control-patterns.md
+++ b/doc/adr/001-version-control-patterns.md
@@ -49,7 +49,7 @@ The first commit should contain the following changes:
   * Delete `templates/banner_html_var.hbs`
   * Put the CTRL text both in `templates/banner_text.hbs` and `templates/banner_text_var.hbs`
   * Reference the correct templates in `banner_var.js`
-If this is a design change test:
+* If this is a design change test:
   * Copy the contents of `banner_ctrl.js` to `banner_var.js`.
   * If this is a markup change, copy the contents of `templates/banner_html.hbs` to `templates/banner_html_var.hbs` and change the template reference in `banner_var.js`  
   * Delete `templates/banner_text_var.hbs` and make sure, `templates/banner_html_var.hbs` includes `templates/banner_text.hbs`
--- a/doc/adr/001-version-control-patterns.md
+++ b/doc/adr/001-version-control-patterns.md
@@ -44,7 +44,7 @@ From time to time we check if the instructions in this document still match our 
 The first commit should contain the following changes:
 * Changes to `campaign_info.toml`, adjusting banner and campaign names.
 * Deleting `styles_var.pcss`
-* Copy the contents of the last "winning" banner to `banner_ctrl.js`, `templates/banner_html.hbs` and `css/styles.styles.pcss`.  
+* Copy the contents of the last "winning" banner to `banner_ctrl.js`, `templates/banner_html.hbs` and `css/styles.pcss`.  
 * If this is a text change test:
   * Delete `templates/banner_html_var.hbs`
   * Put the CTRL text both in `templates/banner_text.hbs` and `templates/banner_text_var.hbs`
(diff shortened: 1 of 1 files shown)


doc/adr/001-version-control-patterns.md
```diff
@@ -0,0 +1,99 @@
...
+
+### Branches
+The `master` branch should contain the banners we have deployed on CentralNotice and the wikipedia.de.
+To develop new banners, we create a feature branch. We merge the feature branches **after** we deployed them to CentralNotice/wikipedia.de.
+Branch names follow the pattern `C<year>_<environment>_<campaign_number>`. `<year>` is the two-digit year number,
+`<environment>` is the environment that displays the banner (`desktop`, `mobile`, `tablet`, `english`, `wpde`).
+`<campaign_number>` is a zero-padded 2-digit number counting up for each A/B test on a platform.
+It should be the [same number used in CentralNotice][1].
+
+If we inadvertently commit something to master, we live with that commit and don't change the master branch.
```
ghost, line 36: Was on the verge of deleting this comment because I feel a bit uncertain after re-reading it - however, I figured I would post it anyway, maybe just so that we can discuss it:

I am not a big fan of rules like that, I think there is no good reason to outright forbid to amend the master branch if we discover a minor mistake **shortly after** merging a banner, at least in the scope of this repository and under the current working mode.

I would argue that due to the fact that essentially only three people are working on the banner code, a quick unanimously-agreed upon exception should be up for discussion and we can all make sure that all master branches are aligned, especially if something was committed to master by mistake.

However, feel free to expand on that rule, maybe I am missing the full picture here.

doc/adr/001-version-control-patterns.md
```diff
@@ -0,0 +1,99 @@
...
+  * Delete `templates/banner_text_var.hbs` and make sure, `templates/banner_html_var.hbs` includes `templates/banner_text.hbs`
+
+The first commit message should have the following format:
+```
+Prepare Campaign <CAMPAIGN_NAME>
+
+Resetting CTRL to (CTRL or VAR from <OLD_CAMPAIGN_NAME>)
+```
+
+[See setup instrcution for campaign names][1].
```
ghost: Typo in instruction

doc/adr/001-version-control-patterns.md
gbirke: Thanks for bringing this up. "Never change master, never rewrite the public history" was an tenet that I somehow picked up during my years as a programmer, but never asked why. For public open source projects it makes sense, because rewriting master might break things for people or automated processes who checked out the old version.

In our case, your argument makes total sense.  It's important that before changing the master all team members have a quick chat to synchronize our versions and make sure that nothing will break. After that, I'm all up for sticking to our commit policies even more. I mostly put this paragraph in to signal "Hey, mistakes happen and it's ok", but having "Hey, mistakes happen and we can fix them" is even better. 👍 🍻 😺

doc/adr/001-version-control-patterns.md
```diff
@@ -0,0 +1,99 @@
...
+### Commits
+Each feature branch should have **two commits**.
+The reviewer of the branch checks if the committer followed the recommendations for the commit contents and format.
+From time to time we check if the instructions in this document still match our banner architecture and update the instructions accordingly.
+
+#### First commit
+The first commit should contain the following changes:
+* Changes to `campaign_info.toml`, adjusting banner and campaign names.
+* Deleting `styles_var.pcss`
+* Copy the contents of the last "winning" banner to `banner_ctrl.js`, `templates/banner_html.hbs` and `css/styles.styles.pcss`.
```
tzhelyazkova: is this just `styles.pcss` ?

gbirke: 👍

PR #126, opened 2018-08-15
C18 WMDE Test 04

ghost:
https://phabricator.wikimedia.org/T201517

### DIFF
--- a/campaign_info.toml
+++ b/campaign_info.toml
@@ -4,7 +4,7 @@
 
 # A Campaign with global settings
 [desktop]
-campaign_tracking = "03-ba-180809"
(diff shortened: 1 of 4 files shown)


ghost:
Don't merge until Campaign is live but feel free to check if everything appears to be as intended. I have changed the way the button text is vertically centered to use flex boxes so if you see any issues with that, let me know.

tzhelyazkova (commented):
Personally, it really bothers me that the buttons' radius remains the same when the viewport changes (when the buttons are vertically aligned). 
I know this is not part of the requirements, therefor we can improve it with the next version, hopefully. :pray:

PR #129, opened 2018-08-20
Prepare Campaign C18_WMDE_Test_mob01

tzhelyazkova:
Resetting `CTRL` to `CTRL` from `C17_WMDE_Test_mob05`


ghost (approved):
This 🆒 commit is 👇 hands 👇 down 👇 🔥 lit 🔥 af 😎.

mobile/banner_ctrl.js
```diff
@@ -80,8 +79,8 @@ $bannerContainer.html( bannerTemplate( {
 	progressBar: progressBar.render()
 } ) );
 
-const trackingEvents = new UrlTracker( trackingBaseUrl, BannerName, $( '.banner-tracking' ) );
-trackingEvents.trackClickEvent( $( '.mini-banner' ), 'banner-expanded' );
+const trackingEvents = new EventLoggingTracker( BannerName );
+trackingEvents.trackClickEvent( $( '.mini-banner' ), 'mobile-mini-banner-expanded' );
```
gbirke, line 83: 👍

mobile/banner_var.js
```diff
@@ -95,19 +94,41 @@ $( '#amount1, #amount2, #amount3, #amount4, #amount5' ).click( function () {
 	$( '#input_amount_other_box' ).val( '' );
 } );
 
+// Disable Sofort-Überweisung payment option if chosen interval is not 'one time'
+$( 'input[name=interval]' ).click( function () {
```
gbirke, line 97: I would like our event handling to be more "expressive", meaning that you don't have to look at the event handling code and the markup to see what's happing.

A first step would be to put the code into a named function, e.g. `toggleIntervalDependentPaymentOptions` and specify that function as the click handler.

An even cooler approach (at least in my opinion) would be to use the DOM event handlers to trigger synthetic events like `interval:selected` on the central banner element and add handlers for those. If you look at `bannerFunctions.js` you can see that some of those already exist. The benefit of those synthetic events are:
- we have some kind of "domain language" of what can happen in a banner. 
- we are more independent from the markup
- we can have a centralized event handling setup function
- the DOM event handlers are short and concise

The existing event names in `bannerFunctions.js` are by no means perfect (I would prefer them to have a real `noun:verb` format). Also, you could argue that it adds an unnecessary layer of indirection. Or that we should use a different, more component-oriented framework/architecture anyway. So feel free to only use function event handlers.

mobile/banner_var.js
```diff
@@ -95,19 +94,41 @@ $( '#amount1, #amount2, #amount3, #amount4, #amount5' ).click( function () {
...
+		$( 'button[data-payment-type=SUB]' ).attr( 'disabled', true ).addClass( 'disabled' );
+	} else {
+		$( 'button[data-payment-type=SUB]' ).attr( 'disabled', false ).removeClass( 'disabled' );
+	}
+} );
+
 $( '.button-group__container button' ).click( function ( event ) {
-	var $checkedAmountElement = $( 'input[name=betrag_auswahl]:checked' );
-	if ( $checkedAmountElement.length > 0 ) {
+	var $checkedAmountElement = $( 'input[name=betrag_auswahl]:checked' ).val(),
```
gbirke, line 106: Please make the event handler a `validateForm` function.

gbirke (approved):
I really like how the 2nd commit makes it easy to see what happens in the `VAR` banner. This made reviewing a breeze!

As the banner functionality is already approved by Kai, I won't withhold my approval, but strongly suggest making the architectural improvements for the event handling code.

tzhelyazkova:
@gbirke thanks for the extensive review :pray: 
I decided to keep the "style" of the code when I implemented the additional click event. I do agree with your remarks and will implement the improvements in a different PR.

PR #131, opened 2018-08-31
Create email subscription box for wikipedia.de

ghost:
T192272

### DIFF
--- a/wikipedia.de_featurebox/banner.js
+++ b/wikipedia.de_featurebox/banner.js
@@ -1,0 +1,62 @@
+require( './css/styles.pcss' );
+
+const bannerTemplate = require( './templates/banner.hbs' );
+const $bannerContainer = $( '.feature' );
+
+$bannerContainer.html( bannerTemplate() );
+
+$( document ).ready( function () {
+	const errorElement = $( '.call-to-action-error' );
+	registerEvents();
+
+	function registerEvents() {
+		let button = $( '#call-to-action-button' );
+
+		$( 'textarea' ).keyup( function ( e ) {
+			if ( e.keyCode === 13 ) {
+				button.trigger( 'click' );
+			}
+		} );
+		button.on( 'click', function () {
+			const email = $.trim( $( '#call-to-action-input' ).val() );
+			if ( isEmail( email ) === false ) {
+				errorElement.text( 'Bitte geben Sie eine gültige E-Mail-Adresse ein.' );
+				return;
+			}
+			$( '#call-to-action-email' ).text( email );
+			$.ajax( {
+				url: 'https://spenden.wikimedia.de/contact/subscribe',
+				data: { email: email },
+				dataType: 'jsonp',
+				jsonpCallback: 'subscriptionCallback'
+			} );
+		} );
+	}
+
+	window.subscriptionCallback = function ( message ) {
+		if ( message.status === 'OK' ) {
+			$( '.call-to-action' ).hide();
+			$( '.call-to-action-confirmation' ).show();
+			return;
+		}
+
+		if ( message.errors.subscription_duplicate_subscription ) {
+			errorElement.text( 'Die angegebene E-Mail-Adresse wurde bereits eingetragen.' );
+			return;
+		}
+
+		errorElement.text( '' );
+		if ( message.errors ) {
+			$.each( message.errors, function ( key, value ) {
+				errorElement.append( value + '</br>' );
+			} );
+			return;
+		}
+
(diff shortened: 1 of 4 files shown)


wikipedia.de_featurebox/banner.js
```diff
@@ -0,0 +1,62 @@
...
+$( document ).ready( function () {
+	const errorElement = $( '.call-to-action-error' );
+	registerEvents();
+
+	function registerEvents() {
+		let button = $( '#call-to-action-button' );
+
+		$( 'textarea' ).keyup( function ( e ) {
+			if ( e.keyCode === 13 ) {
+				button.trigger( 'click' );
```
gbirke: I can't really say why, but I generally try to avoid triggering click events on other elements. I'd rather find a common event handling function.

wikipedia.de_featurebox/banner.js
```diff
@@ -0,0 +1,62 @@
...
+			}
+		} );
+		button.on( 'click', function () {
+			const email = $.trim( $( '#call-to-action-input' ).val() );
+			if ( isEmail( email ) === false ) {
+				errorElement.text( 'Bitte geben Sie eine gültige E-Mail-Adresse ein.' );
+				return;
```
gbirke: Just returning won't cancel the default event handling of the browser. A better alternative is to capture the venet as function parameter and call `event.preventDefault()`. Whith jQuery you can also return `false`, which will also stop the event from bubbling up.

wikipedia.de_featurebox/banner.js
```diff
@@ -0,0 +1,62 @@
...
+			}
+			$( '#call-to-action-email' ).text( email );
+			$.ajax( {
+				url: 'https://spenden.wikimedia.de/contact/subscribe',
+				data: { email: email },
+				dataType: 'jsonp',
+				jsonpCallback: 'subscriptionCallback'
```
gbirke: I reaserached some more and found out that it also works by omitting the `jsonpCallback` property and providing a `success` callback instead.

wikipedia.de_featurebox/banner.js
```diff
@@ -0,0 +1,62 @@
...
+				errorElement.append( value + '</br>' );
+			} );
+			return;
+		}
+
+		errorElement.text( 'Beim Eintragen der Adresse ist ein Fehler aufgetreten.' );
+	};
+
+	function isEmail( email ) {
+		return /^(([a-zA-Z0-9_.-])+@([a-zA-Z0-9_.-])+\.([a-zA-Z])+([a-zA-Z])+)?$/.test( email );
```
gbirke: Ever since I read https://hackernoon.com/the-100-correct-way-to-validate-email-addresses-7c4818f24643 I prefer the client-side validation for email addresses to be as lenient as possible, basically checking if there is an `@` in there after the 1st char.

Bad email validation has become some pet peeve of mine, since many website make simplistic assumptions. As some person on Stackoverflow said, "[You have to worry about 3-4 RFCs to do it correctly](https://stackoverflow.com/a/12355897/130121)"
* Some mail providers (e.g. Gmail) have this nifty feature of adding a `+prefix` before the `@` sign, providing you with unlimited email addresses that allow you to track data leaks. But you have to allow `+` in mail addresses and many validations on the web don't. 
* Another thing many validators get wrong (I think even parts of FundraisingFrontend might be guilty) is international domains. Got lieschen@müller.de? Good luck providing your mail address.

Also, your regex allows empty strings, I guess that's not intentional?

wikipedia.de_featurebox/templates/banner.hbs
```diff
@@ -0,0 +1,22 @@
...
+        <p><input id="call-to-action-input" type="email"/></p>
+        <p class="call-to-action-error"></p>
+        <p class="call-to-action-button-wrapper">
+            <button id="call-to-action-button" type="button">E-Mail eintragen</button>
+        </p>
+    </div>
+
+    <div class="call-to-action-confirmation" style="display:none;">
+        <p>
+            Sie werden ab sofort über E-Mail an <span id="call-to-action-email">die von Ihnen angegebene E-Mail-Adresse</span> informiert, wenn Wikipedia Ihre Hilfe benötigt.
```
gbirke: I've asked in the ticket about a different text, since we do double-opt-in.

gbirke (changes requested):
I would like the code to have more structure, to be more readable. To have some separation between
* handling DOM events, 
* receiving and processing server data,
* updating the DOM, 
* etc

I'm fine with whatever method you choose - named functions, event bus, using Vue.js, etc. I'll do a followup with a suggestion using named functions.

PR #132, opened 2018-08-31
C18 WMDE Test 05

ghost:
On the variant banner, the "Lastschrift" payment method uses a different URL and some different data so it's important to differentiate between that and the other payment methods when testing. What's weird to me is that da=1 seems to be broken / lost, any idea on why that could be @tzhelyazkova @gbirke?

### DIFF
--- a/desktop/banner_var.js
+++ b/desktop/banner_var.js
@@ -86,6 +86,8 @@ $bannerContainer.html( bannerTemplate( {
 	progressBar: progressBar.render()
 } ) );
 
+const $bannerForm = $( '#WMDE_Banner-form' );
+
 // BEGIN form init code
 
 const trackingEvents = new EventLoggingTracker( BannerName );
@@ -135,6 +137,7 @@ function setupAmountEventHandling() {
 	banner.on( 'amount:selected', null, function () {
 		$( '#amount-other-input' ).val( '' );
 		$( '.select-group__custom-input' ).removeClass( 'select-group__custom-input--value-entered' );
+		$( '#betrag' ).val( BannerFunctions.getAmount() );
 		BannerFunctions.hideAmountError();
 	} );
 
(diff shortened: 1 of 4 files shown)


desktop/templates/banner_html_var.hbs
```diff
@@ -9,7 +9,9 @@
 
             <div class="form">
                 <form id="WMDE_Banner-form" method="post" name="donationForm"
-                      action="https://spenden.wikimedia.de/donation/new?piwik_campaign={{ CampaignName }}&piwik_kwd={{ BannerName }}">
+                      action=""
+                      data-main-action="https://spenden.wikimedia.de/donation/add?piwik_campaign={{ CampaignName }}&piwik_kwd={{ BannerName }}&da=1&mbt=1"
+                      data-fallback-action="https://spenden.wikimedia.de/donation/new?piwik_campaign={{ CampaignName }}&piwik_kwd={{ BannerName }}&da=0">
```
gbirke, line 14: AFAIR using da=0 in VAR will skew the results and it still should be 1, even when the behavior does not change for direct debit.
Right, @KaiNissen ?

KaiNissen: I think, direct debit payments will be ignored anyway when analyzing. So it does not really matter whether these donations log as the variant.

desktop/banner_var.js
```diff
@@ -143,10 +146,16 @@ function setupAmountEventHandling() {
 		var input = $( '.select-group__custom-input' );
 		input.addClass( 'select-group__custom-input--value-entered' );
 		BannerFunctions.hideAmountError();
+		$( '#betrag' ).val( BannerFunctions.getAmount() );
 		appendEuroSign( input );
 	} );
 
 	banner.on( 'paymenttype:selected', null, function () {
+		let bannerAction = $bannerForm.data( 'main-action' );
+		if ( $( 'input[name=zahlweise]:checked', document.donationForm ).val() === 'BEZ' ) {
```
gbirke, line 155: Why not use `$bannerForm.find('input[name=zahlweise]:checked')` or `$('input[name=zahlweise]:checked', $bannerForm)`? referencing document.something is so 90s

gbirke: Apart from the fact I always feel sightly dirty nowadays when "application state" is kept in the DOM, but that's a whole other discussion and definitely not part of this PR.

gbirke (approved):
I've user-tested the banners (on my local machine) and they work.

desktop/banner_var.js
ghost: @gbirke I got this by looking at the shared banner functions:

https://github.com/wmde/fundraising-banners/blob/b5ba64390ee1f2c91b1ff6486e345679c0e9ad72/shared/banner_functions.js#L204

I was also wondering about what the point of it was and figured I would rather leave it that way in case this was done intentionally and was scared that it could break something if I changed it, knowing you would surely comment on it if it was indeed odd :nerd_face:. I guess we should change both instances then before the go-live?

PR #137, opened 2018-09-13
Fix amount handling

gbirke:
As different end points expect different locales, we need to format the
locales whenever the payment type (which determines the end point)
changes.

See https://phabricator.wikimedia.org/T204202

This solves https://phabricator.wikimedia.org/T204113


gbirke:
The fix is already deployed because it was an urgent bugfix. Please review and make changes as needed.

PR #139, opened 2018-09-25
Mobile Test 02 / 2018

ghost:
https://phabricator.wikimedia.org/T205108

There is still no text for the individual mini banner slides at this time which will have to be added later, the rest of the banner should be done though.

### DIFF
--- a/mobile_slider/banner_ctrl.js
+++ b/mobile_slider/banner_ctrl.js
@@ -15,7 +15,6 @@ const BannerFunctions = require( '../shared/banner_functions' )( GlobalBannerSet
 const formatNumber = require( 'format-number' );
 const CampaignProjection = require( '../shared/campaign_projection' );
 const bannerTemplate = require( './templates/banner_html.hbs' );
-const ProgressBar = require( '../shared/progress_bar/progress_bar' );
 
 require( './css/styles.pcss' );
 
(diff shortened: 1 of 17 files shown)


mobile_slider/banner_ctrl.js
```diff
@@ -62,17 +61,7 @@ const weekdayPrepPhrase = dayName.isSpecialDayName() ? Translations[ 'day-name-p
 const $bannerContainer = $( '#WMDE-Banner-Container' );
 const CampaignName = $bannerContainer.data( 'campaign-tracking' );
 const BannerName = $bannerContainer.data( 'tracking' );
-const numberOfDaysUntilCampaignEnd = campaignDays.getNumberOfDaysUntilCampaignEnd();
```
tzhelyazkova, line 65: Why is this removed?

mobile_slider/banner_ctrl.js
```diff
@@ -91,9 +92,11 @@ $( '.select-group__option' ).click( function () {
 $( 'input[name=interval]' ).click( function () {
 	const subPaymentButton = $( 'button[data-payment-type=SUB]' );
 	if ( $( this ).attr( 'id' ) !== 'interval_onetime' ) {
+		if ( subPaymentButton.hasClass( 'selected-option' ) ) {
```
tzhelyazkova, line 92: That check is probably not needed, because `removeClass()` won't return an error or do something unexpected when the class is not there.

mobile_slider/banner_ctrl.js
```diff
@@ -118,6 +121,22 @@ $( '.payment-selection button' ).click( function ( event ) {
 	$( '#zahlweise' ).val( $( event.target ).data( 'payment-type' ) );
 } );
 
+$( '#banner-form-submit' ).click( function () {
+	if ( !$( '#periode' ).val() || jQuery( '.interval-selection .selected-option' ).length === 0 ) {
```
tzhelyazkova: Is there a reason for using `jQuery` instead of `$` ?

tzhelyazkova (commented):
I guess what's left is to split the code into var and ctrl parts.

mobile_slider/banner_ctrl.js
ghost: If it's not there, `$( '#zahlweise' ).val( '' );` will always reset the input field value when Sofortüberweisung is disabled. With the check in place, it will only do that if Sofortüberweisung was previously selected. I should probably add a comment that clarifies this though.

mobile_slider/banner_ctrl.js
ghost: Nah, I goofed after checking some stuff in the JavaScript console.

mobile_slider/banner_ctrl.js
ghost: It was only needed for the progress bar, it will be added back alongside the progress bar later during the campaign.

tzhelyazkova:
Tried it on Safari, it looks pretty ok to me. 
Behaves as described in the ticket, besides the clicking on everywhere on the small banner to open the form. But that I assume would be impossible for now.

tzhelyazkova (approved):

ghost:
@tzhelyazkova @gbirke Please check 3d362cb and the corresponding changes I have made to the BannerEvents:

https://meta.wikimedia.org/w/index.php?title=Schema:WMDEBannerEvents&oldid=18437830

tzhelyazkova:
The changes seem alright to me :footprints:

tzhelyazkova:
Merged this because it went live and the new `EventLogging` schema was also merged and deployed to production

PR #140, opened 2018-09-28
Banner Test 08

ghost:


ghost:
The list of requirements for the variant banner is relatively large, so maybe see if I missed anything:

https://phabricator.wikimedia.org/T205115

desktop/templates/banner_html_var.hbs
```diff
@@ -156,8 +156,7 @@
 
         <div class="close">
             <a href="#" class="close__link">
-                <img src="//upload.wikimedia.org/wikipedia/commons/c/cb/Close_x_-_black.png" alt="schließen"
-                     title="schließen">
+                &#x2715;
```
tzhelyazkova, line 159: clever :+1:

shared/messages/en.js
```diff
@@ -1,9 +1,9 @@
 const Translations = {
```
tzhelyazkova, line 1: Hm, I'm not even sure why we have English translations. Those don't seem to be used anywhere.
Also I don't think it makes sense to use them  since the fundraising application is German only.

desktop/templates/banner_text_var.hbs
```diff
@@ -0,0 +1,25 @@
...
+	Millionen Menschen nutzen Wikipedia, aber nur ein Bruchteil spendet.
+
+	<span class="text__paragraph_screen_l">Als ich mich entschloss, Wikipedia zu einem nicht-kommerziellen Projekt zu machen, wurde mir gesagt, ich würde dies bereuen.</span>
+
+	<span class="text__paragraph_screen_m">Sicher könnten wir mit Werbung eine Menge Geld verdienen. Aber dann wäre Wikipedia komplett anders. Wir könnten ihr nicht vertrauen.</span>
+
+	<span class="text__paragraph_screen_m">Es ist leicht, diese Nachricht zu ignorieren und die meisten werden das wohl tun.</span>
+
+	Wenn Sie Wikipedia nützlich finden, nehmen Sie sich {{ weekdayPrepPhrase }} {{ currentDayName }}
+	bitte eine Minute Zeit und geben Wikipedia mit Ihrer Spende etwas zurück. Vielen Dank.
```
tzhelyazkova: In the banners of Campaign 07, the sentence ends with exclamation point rather than a full stop. I think that should be the case here as well.

tzhelyazkova (changes requested):
- The `us` url parameter needs to be removed as it's part of the currently ongoing campaign (06 I think)
- When entering custom amount the euro sign is appended and then disappears. :cloud:  
- The sentence `"Heute ist sie die einzige Top-10-Website, die gemeinnützig betrieben wird."` is missing from the `ctrl` banner.
- On small viewports (< 800) in portrait mode, the form appears to not be aligned on the left, or centered. I think if the margin of `.form` is set to `0` it will be in alignment with the infobox.
- The sentence `Es ist leicht, diese Nachricht zu ignorieren und die meisten werden das wohl tun.` is missing from the `800-960` viewport

shared/messages/en.js
ghost: 🤷‍♂️

ghost: @KaiNissen Might know more.

tzhelyazkova (approved):

shared/messages/en.js
KaiNissen: We are running banners on enwiki and Fundraising Ops also considers to run banners on mobile enwiki and Commons this year.

PR #142, opened 2018-10-16
Implement randomized banner viewport tracking

ghost:


shared/event_logging_tracker.js
```diff
@@ -50,11 +72,19 @@ export default class EventLoggingTracker {
...
 	}
 
-	track( actionName, trackingRatio, slidesShown, finalSlide ) {
+	/**
+	 * Track banner event data
+	 * @param {string} actionName
+	 * @param {number} trackingRatio
+	 * @param {number} slidesShown
+	 * @param {number} finalSlide
+	 * @private
```
gbirke: Is this really private?

shared/event_logging_tracker.js
```diff
@@ -50,11 +72,19 @@ export default class EventLoggingTracker {
...
+	 */
+	trackBannerEvent( actionName, trackingRatio, slidesShown, finalSlide ) {
```
gbirke: Yay for renaming and adding doc blocks!

shared/track_size_issues.js
```diff
@@ -30,7 +30,7 @@ export default class SizeIssueIndicator {
 	}
 
 	/**
-	 * Check if Banner takes too much screen space and track the incident
+	 * Check if Banner takes too much screen space and trackBannerEvent the incident
```
gbirke: [A clbuttic mistake](https://www.telegraph.co.uk/news/newstopics/howaboutthat/2667634/The-Clbuttic-Mistake-When-obscenity-filters-go-wrong.html) 😉

desktop/banner_ctrl.js
```diff
@@ -294,6 +294,11 @@ $( function () {
 		return;
 	}
 
+	trackingEvents.trackViewPortDimensions(
```
gbirke, line 297: We'll have to ask Kai if the data should be tracked even if the banner is not shown. It's kind of double tracking otherwise.

shared/event_logging_tracker.js
```diff
@@ -22,13 +22,35 @@ export default class EventLoggingTracker {
...
+	 */
+	trackViewportData( dimensionData, trackingRatio, bannerPrefix ) {
 		if ( typeof trackingRatio === 'undefined' ) {
 			trackingRatio = 1;
 		}
 
 		if ( Math.random() < trackingRatio ) {
 			mw.track( 'event.WMDEBannerSizeIssue', {
-				bannerName: this.bannerName,
+				bannerName: bannerPrefix + this.bannerName,
```
gbirke: We need to discuss with Kai if we're interested in the overall screen size data across many banners and campaigns or individual campaigns/banners. 

 If we don't care about individual campaigns/banners then it would make sense to use an identifier instead of prefix+banner name.

In the unlikely case that we want to query the screen sizes for individual campaigns/banners, then it would be better to use a suffix than a prefix: When doing the export query with a `LIKE` statement, the database can use an index when the query does not start with a wildcard.

shared/event_logging_tracker.js
```diff
@@ -22,13 +22,35 @@ export default class EventLoggingTracker {
...
+	}
+
+	/**
+	 * Track the dimensions of the user viewport for statistical purposes
+	 *
+	 * @param {object} dimensionData
+	 * @param {number} trackingRatio The probability of the event being tracked (between 0 and 1)
+	 */
+	trackViewPortDimensions( dimensionData, trackingRatio ) {
+		this.trackViewportData( dimensionData, trackingRatio, 'viewport_tracking_' );
```
gbirke: Maybe use a clearer separator like `/` or `#` between prefix and banner name. Otherwise you get this weird mix of snake_and-kebab-case

gbirke (commented):
+1 for the implementation, let's ask Kai for the missing information.

shared/event_logging_tracker.js
```diff
@@ -22,13 +22,35 @@ export default class EventLoggingTracker {
...
+	 *
+	 * @param {object} dimensionData
+	 * @param {number} trackingRatio
+	 * @param {string} bannerPrefix
+	 * @private
+	 */
+	trackViewportData( dimensionData, trackingRatio, bannerPrefix ) {
 		if ( typeof trackingRatio === 'undefined' ) {
 			trackingRatio = 1;
 		}
```
tzhelyazkova: I don't quite understand the above `if statement`.
First, if `trackingRatio` is `undefined` and then given value of `1`, then `Math.random() < trackingRatio` will always be true, therefor we track 100%. 
But what is the case in which trackingRatio is `undefined`? If it's meant to check for a possible incorrect call of the function, then we shouldn't default to tracking 100%. If it's meant as in "this is how you call the function if you want to track 100%" then I believe we should make it more explicit and have the function be called with `trackingRatio` of `1`.
`typeof` is considered "not safe" since the introduction of `let` and `const`. I think a better approach would be to always pass a value of `trackingRatio`, and then for extra safety make the check `if ( Math.random() < Number( trackingRatio ) )` which will throw and exception if `trackingRatio` is `undefined`.
I know this is out of the scope of this PR, but I think it's something that certainly needs change. Tagging @gbirke because `git blame` showed his name :)
p.s. @timEulitz now that you've split the big tracking function into smaller ones (:+1:) not all of them do this check for `trackingRatio` being `undefined` => inconsistency

shared/event_logging_tracker.js
```diff
@@ -22,13 +22,35 @@ export default class EventLoggingTracker {
...
+		this.trackViewportData( dimensionData, trackingRatio, 'viewport_tracking_' );
+	}
+
+	/**
+	 * Track dimension data depending on tracking ratio
+	 *
+	 * @param {object} dimensionData
+	 * @param {number} trackingRatio
+	 * @param {string} bannerPrefix
+	 * @private
```
tzhelyazkova, line 46: probably a copy/paste mistake? Or you meant a private function is returned? But that's also not the case. :thinking:

tzhelyazkova (approved):

PR #147, opened 2018-11-01
C18 WMDE Test 11

ghost:
Potentially still two open questions in https://phabricator.wikimedia.org/T208173


desktop/banner_var.js
```diff
@@ -253,10 +254,13 @@ function displayBanner() {
 	bannerElement.css( 'display', 'block' );
 	addSpace();
```
gbirke: `addSpace` is meant to animate pushing down skin-specific elements to make room for the slide-in banner. In this banner, the space is already reserved by the instantBanner (Which should be exactly the same height as the main banner). So I'd remove this call for performance & clarity reasons.

desktop/templates/banner_html_var.hbs
```diff
@@ -1,10 +1,44 @@
...
+
 <div id="WMDE_Banner">
     <div class="banner">
         <div class="banner__content">
             <div class="infobox">
                 <div class="infobox__text">
-                    {{> banner_text_var }}
+                    {{> banner_text }}
                 </div>
+                {{{ progressBar }}}
```
gbirke, line 47: The 2 progress bars and their animation are not working right at the moment, we need a clear spec from Kai how they should work & adapt the code.

My suggestion: Animate the progress bar in the instant banner immediately when it's shown & set the width in the main banner (without animation).

At the moment, the animation of the progress bar is delayed by 1 second (which is too long IMHO). The main banners progress bar does not animate, but counts up the numbers.

My test settings in global_banner_settings:
```javascript
module.exports = {
	'goalSum': 7900000,
	'donations-date-base': '2018-10-01',
	'donations-collected-base': 10000,
	'donators-base': 0,
	'appr-donations-per-minute': 104,
	'appr-donators-per-minute': 4.805,
	'milion-impressions-per-day': 6, /** @deprecated kept for legacy reasons */
	'impressions-per-day-in-million': 6,
	'allBannersCookie': 'banner_impression_count',
	'singleBannerCookie': 'single_banner_impression_count',
	'campaign-start-date': '2018-10-01',
	'campaign-end-date': '2018-12-31',
	'banner-close-track-ratio': 1
};
```

gbirke (changes requested):

desktop/templates/banner_html_var.hbs
KaiNissen: Sorry for keeping you waiting on this. What Gabriel says, seems reasonable to me: Instantly starting the animation of the progress bar on the "instant banner". The progress bar on the second banner should neither animate nor count up the numbers.

ghost: For the record, this was fixed by removing an animate() call to the progress bar which was there by mistake.

PR #149, opened 2018-11-05
Remove mobile slider prototype (now mobile banner)

ghost:


tzhelyazkova (approved):
So the mobile banner becomes the one with the slider for EVA? :tada: :tada:

campaign_info.toml
```diff
@@ -149,16 +149,4 @@ wrapper_template = "wikipedia_org_js_banner_values"
...
-campaign_tracking = ""
-preview_link = "/wiki/Wikipedia:Hauptseite?useskin=minerva&banner=B17WMDE_webpack_prototype&devbanner={{banner}}"
-wrapper_template = "wikipedia_org_js_banner_values"
-
-[mobile_slider.banners.ctrl]
-filename = "./mobile_slider/banner_ctrl.js"
-pagename = "B18WMDE_mob_prototype"
-tracking = "org-mob06-180501-ctrl"
\ No newline at end of file
+tracking = "org-desktop-en-171121"
```
KaiNissen:
```suggestion
tracking = "org-desktop-en-171121"

```

Not a real issue, but I think at some point we agreed that files should always end with an empty line. Not sure, if this also applies to toml, though.

PR #168, opened 2018-11-15
C18_WMDE_Test_13

tzhelyazkova:
[T209595](https://phabricator.wikimedia.org/T209595)


ghost (approved):
I guess we'll see how successful this change is when you have to either lower or increase the average tomorrow 💰.

PR #182, opened 2018-11-30
C18 WMDE Test 19

ghost:
New banner fresh off the press 📰!


desktop/css/styles.pcss
```diff
@@ -144,7 +144,7 @@
 	}
 
 	.application-of-funds-link {
-		color: var( --banner-color-gray );
+		color: var( --banner-color-black );
```
tzhelyazkova, line 147: The bank data line text should also be set to `--banner-color-black`.

PR #191, opened 2018-12-05
C18_WMDE_Test_21

tzhelyazkova:
[T211081](https://phabricator.wikimedia.org/T211081)
IE = :hankey:


tzhelyazkova:
The CI will keep failing because the `js` refers to html elements that have been removed.

PR #192, opened 2018-12-06
Resolve non-breaking security issues via npm audit

ghost:


ghost:
Everything seems to still work fine after this.

package.json
```diff
@@ -37,7 +37,7 @@
     "minimatch": "^3.0.4",
     "mocha": "^5.2.0",
     "mocha-jsdom": "^1.1.0",
-    "postcss-custom-properties": "^7.0.0",
```
ghost, line 40: This was not strictly needed but is just an extra update.

gbirke (commented):
The npm version from our Docker image, digitallyseamless/nodejs-bower-grunt, does not support the audit command. To avoid us running into trouble with with different local npm versions, I'll modify this PR with switching the docker image to the latest node LTS version.

PR #193, opened 2018-12-06
Fix missing compiled variables in mini banner

ghost:
Due to the variables not being in the same file, postcss fails to load
them and does not replace the var() blocks with backwards compatible
raw values. By forcing postcss to not preserve the original variables, we
reduce the size of the output CSS and make sure that we notice this more
easily, as missing variables would now break on all browsers.

https://phabricator.wikimedia.org/T211099

### DIFF
--- a/mobile/css/styles_mini.pcss
+++ b/mobile/css/styles_mini.pcss
@@ -1,3 +1,18 @@
+:root {
+	--banner-color-red: #990a00;
+	--banner-color-red-light: #dbb0ad;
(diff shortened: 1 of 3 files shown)


ghost:
@gbirke @tzhelyazkova This fix worked for the mobile banners on Kai's phone and desktop banner 21, so I would say we merge this and keep in mind that we always need to define all relevant CSS variables in each file.

tzhelyazkova (approved):

PR #194, opened 2018-12-06
Fix lateprogress progress bar style hack

ghost:
https://phabricator.wikimedia.org/T211350


gbirke (commented):
Now the lateprogress modifier does not look good :(

Let's discuss in the next daily if we want to maintain 2 versions of the progress bar throughout the tests.

PR #201, opened 2018-12-10
Include tracking data in WPDE banners via webpack

ghost:


webpack.common.js
```diff
@@ -66,6 +67,13 @@ module.exports = {
 			},
 			filePattern: 'B*.js',
 			campaignConfig: campaigns.getConfigForPages()
+		} ),
+		new WrapperPlugin( {
+			test: /B18WPDE.*.js$/,
```
tzhelyazkova: This means we will have to change it every year :question: 
Suggestion: `B.\dWPDE.*.js$`

webpack.common.js
```diff
@@ -66,6 +67,13 @@ module.exports = {
...
+			header: function ( pageName ) {
+				const trackingData = campaigns.getCampaignTracking( pageName.replace( '.js', '' ) );
+				return `var BannerName = '${trackingData.bannerTracking}'; var CampaignName = '${trackingData.campaignTracking}';`;
```
tzhelyazkova, line 75: neat ES6 magic

webpack.common.js
```diff
@@ -66,6 +67,13 @@ module.exports = {
 			},
 			filePattern: 'B*.js',
 			campaignConfig: campaigns.getConfigForPages()
+		} ),
+		new WrapperPlugin( {
+			test: /B\d{2}WPDE.*.js$/,
```
tzhelyazkova, line 72: 82 years from now some :alien: entity :alien: will spend lots of time investigating why their banners don't show on `wikipedia.de` and will curse the ancient devs :older_man: :older_woman:

tzhelyazkova (approved):

PR #202, opened 2018-12-11
Collect viewport size on close button clicks

ghost:
https://phabricator.wikimedia.org/T211662

### DIFF
--- a/desktop/banner_ctrl.js
+++ b/desktop/banner_ctrl.js
@@ -236,10 +236,6 @@ $( '#impCount' ).val( impCount );
 var bannerImpCount = BannerFunctions.increaseBannerImpCount( BannerName );
 $( '#bImpCount' ).val( bannerImpCount );
 
-// track lightbox link clicking and banner closing
-trackingEvents.trackClickEvent( $( '#application-of-funds-link' ), 'application-of-funds-shown', 1 );
-trackingEvents.trackClickEvent( $( '#WMDE_Banner .close__link' ), 'banner-closed', bannerCloseTrackRatio );
-
 // BEGIN Banner close functions
 $( '#WMDE_Banner .close__link' ).click( function () {
 	$( '#WMDE_Banner' ).hide();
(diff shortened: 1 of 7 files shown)


desktop/banner_ctrl.js
```diff
@@ -262,13 +258,24 @@ $( '#ca-ve-edit, .mw-editsection-visualeditor' ).click( function () {
...
 
 	$( 'body' ).prepend( $( '#centralNotice' ) );
 
 	if ( BannerFunctions.onMediaWiki() && window.mw.config.get( 'wgAction' ) !== 'view' ) {
 		return;
 	}
 
+	// track lightbox link clicking and banner closing
+	trackingEvents.trackClickEvent( $( '#application-of-funds-link' ), 'application-of-funds-shown', 1 );
+	trackingEvents.trackClickEvent( closeLink, 'banner-closed', bannerCloseTrackRatio );
```
gbirke: You need to remove this line in all the banners, as the closing event is sent inside `trackCloseEventViewPortDimensions`. Then you can also get rid of the `closeLink` variable again.

shared/event_logging_tracker.js
```diff
@@ -37,6 +38,24 @@ export default class EventLoggingTracker {
...
+	 * @param {jQuery} $trackedElement
+	 * @param {function} dimensionCallback
+	 * @param {number} slidesShown
+	 * @param {number} finalSlide
+	 * @param {number} trackingRatio The probability of the event being tracked (between 0 and 1)
+	 */
+	trackCloseEventViewPortDimensions( $trackedElement, dimensionCallback, slidesShown, finalSlide, trackingRatio = 0.01 ) {
+		$trackedElement.click( function () {
+			if ( this.trackBannerEvent( 'banner-closed', slidesShown, finalSlide, trackingRatio ) ) {
+				// 10% chance to also track viewport dimensions on top of recording the banner closed event.
```
gbirke, line 53: I know Tobias wanted it like that, but can you briefly explain the rationale behind making it not 100%? Is there so much close data that we can get meaningful results by only sampling 10%?

gbirke (changes requested):

shared/event_logging_tracker.js
ghost: Like you said yourself, Tobias asked me for this exact number and since he was not reachable anymore afterwards, I implemented it like that. I don't think the amount of data would be crazy but generally I am in favor of collecting as little as needed. We can double-check with him if it's of concern.

desktop/banner_ctrl.js
ghost: True, done.

shared/event_logging_tracker.js
ghost: @gbirke Updated to 100% now as discussed via email.

gbirke (approved):

PR #204, opened 2018-12-12
C18_WMDE_EN_Test_05

ghost:
https://phabricator.wikimedia.org/T211658


english/banner_var.js
```diff
@@ -1,3 +1,4 @@
+require( './css/styles.pcss' );
```
gbirke, line 1: Did that style not get totally in the way? While both style sheets share some similarity, the markup and design are also different enough that you have to adjust lots of stuff. It does not matter for performance and file size because of we have the `postcss-combine-duplicated-selectors` plugin (although styles.pcss might contain unused selectors), but I wonder if importing this made your work unnecessarily complicated.

PR #207, opened 2018-12-13
C18 WMDE iPad Test 04

tzhelyazkova:


shared/progress_bar/style.pcss
```diff
@@ -87,7 +87,7 @@
 	}
 }
 
-@media ( max-width: 900px ) {
```
tzhelyazkova, line 90: This is changed due to progress bar's left side text disappearing on small iPad sizes.
Therefor now this text should only be hidden for mobile. This might be changed in the future.

PR #209, opened 2018-12-14
Fix IE11 wrapping issues on lower resolutions

ghost:
https://phabricator.wikimedia.org/T211936

### DIFF
--- a/desktop/css/styles.pcss
+++ b/desktop/css/styles.pcss
@@ -50,7 +50,8 @@
 
 	.infobox {
 		display: flex;
-		flex-direction: column;
+		flex-direction: row;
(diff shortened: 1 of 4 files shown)


gbirke (approved):
I'm not happy with this "hack", which in my opinion misuses flexbox features and makes the CSS hard to understand. However, it works and I see that you have deeply researched this and found no other solution. Therefore: Thanks for fixing this!

tzhelyazkova (approved):
IE = 💩💩💩
We need to test if the iPad is still ok after these changes. The code on master is test `04`, and the changes here are on files from test `03`. I think it's easier to merge into master and then check compared to the other way around.

PR #214, opened 2018-12-16
Progress bar on mobile banner

tzhelyazkova:
Bug: [T211672](https://phabricator.wikimedia.org/T211672)
The text is moved out from the progress bar to the area above it.
It's not super ideal. Still breaks in a second row for `screen width < 335px`

Possible improvements: 
- Shorten the text that says `N days left` 
- Decrease text size to `12px`
- Change `font-style` to normal, instead of bold

### DIFF
--- a/shared/progress_bar/style_mobile.pcss
+++ b/shared/progress_bar/style_mobile.pcss
@@ -1,10 +1,32 @@
 .progress_bar__text__above {
 	font-size: 14px;
 	font-weight: bold;
-	line-height: 28px;
+	line-height: 18px;
 	display: flex;
-	flex-direction: row;
-	justify-content: space-around;
-	flex-wrap: nowrap;
+	flex-wrap: wrap;
 	padding: 0 20px;
+
+	@media ( max-width: 350px ) {
+		font-size: 13px;
+	}
+
+	> div {
+		white-space: nowrap;
+	}
+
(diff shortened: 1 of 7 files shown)


ghost:
I have pushed some more improvements which should resolve the issues for lower resolutions. Before merging we should clean up the old progress bar code and use this for all banners but I will double-check with Kai on that.

gbirke (changes requested):
The right text looks a bit weird, as if a separator was missing. My proposal:

> Nur noch x Tage, es fehlen noch y Mio."

We should reintroduce the rounded corners of the progress bar, at the moment it looks a bit "cut off".

shared/progress_bar/style_mobile.pcss
```diff
@@ -1,10 +1,32 @@
 .progress_bar__text__above {
 	font-size: 14px;
 	font-weight: bold;
-	line-height: 28px;
+	line-height: 18px;
 	display: flex;
-	flex-direction: row;
-	justify-content: space-around;
-	flex-wrap: nowrap;
+	flex-wrap: wrap;
```
tzhelyazkova: hm, why? That will make the text split in two rows which is not what we want, right? (at least according to the mock)

gbirke (approved):

PR #217, opened 2018-12-19
Fix banner height tracking for close events

ghost:
Since the banner height is captured after the banner is hidden, it is
currently tracked as 0. This change switches the event order around
which solves this problem and ensures the correct number is tracked.

### DIFF
--- a/shared/event_logging_tracker.js
+++ b/shared/event_logging_tracker.js
@@ -51,7 +51,7 @@ export default class EventLoggingTracker {
 		$trackedElement.click( function () {
 			if ( this.trackBannerEvent( 'banner-closed', slidesShown, finalSlide, trackingRatio ) ) {
 				// 10% chance to also track viewport dimensions on top of recording the banner closed event.
-				this.trackViewportData( VIEWPORT_TRACKING_CLOSED_EVENT_IDENTIFIER + '-' + this.bannerName, dimensionCallback(), 0.1 );
+				this.trackViewportData( VIEWPORT_TRACKING_CLOSED_EVENT_IDENTIFIER + '-' + this.bannerName, dimensionCallback(), 1 );
 			}
 		}.bind( this ) );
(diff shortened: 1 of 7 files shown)


shared/event_logging_tracker.js
```diff
@@ -51,7 +51,7 @@ export default class EventLoggingTracker {
 		$trackedElement.click( function () {
 			if ( this.trackBannerEvent( 'banner-closed', slidesShown, finalSlide, trackingRatio ) ) {
 				// 10% chance to also track viewport dimensions on top of recording the banner closed event.
-				this.trackViewportData( VIEWPORT_TRACKING_CLOSED_EVENT_IDENTIFIER + '-' + this.bannerName, dimensionCallback(), 0.1 );
+				this.trackViewportData( VIEWPORT_TRACKING_CLOSED_EVENT_IDENTIFIER + '-' + this.bannerName, dimensionCallback(), 1 );
```
tzhelyazkova, line 54: If this is only about fixing the bug why switch the tracking ratio?

ghost: Actually it was discussed as 1 instead of 0.1, I think this commit got lost "somehow". On the liver banners the value 1 was used as well, I guess I forgot to commit this or did not get pushed properly and I did not notice it before. Not 100% clean I guess, if it bugs you that much I can split it out.

tzhelyazkova: :no_entry: Actually :no_entry:  
No it's fine. I was genuinely asking cause there's no ticket to refer to, didn't know if someone asked for this or you decided to bump it up to make up for all the data we missed to track so far.

tzhelyazkova (approved):

shared/event_logging_tracker.js
ghost: :pray:

PR #219, opened 2018-12-20
C18_WMDE_ThankYou

tzhelyazkova:
Thank You banner to encourage people to become a member or help edit Wikipedia
Ticket: [T212324](https://phabricator.wikimedia.org/T212324)


ghost:
@gbirke @tzhelyazkova I just wanted to drop some general feedback before I forget:

I have noticed that the (P)CSS for this banner had multiple cases where the CSS was not structured in a "mobile-first" manner, this lead to some problems because of how PCSS is compiled into CSS. For example, in this case, the width of my window was below 600px:

untitled

Yet, as you can see, the rule for `max-width: 900px` is active instead of the rule for `max-width: 600px` because PCSS compiles all rules and places them in order of their width in the final CSS. This means that the `max-width: 900px` rule is below `max-width: 600px` in the final CSS and takes priority, which means that the rule for `max-width: 600px` will never be active.

The solution is to generally always use min-width, starting with the lowest mobile resolution without any media query and then working up to the highest desktop-resolution with min-width media queries. I am actually almost inclined to say that using `max-width` media queries might be a "code smell" which we should look out for in the future.

ghost:
@gbirke @tzhelyazkova I just played around a little more and what I wrote in my previous comment is not entirely correct. PCSS seems to order its media queries simply by the order it finds media query blocks and it goes through the PCSS by depth to find them (as opposed to reading the PCSS file top to bottom and ordering the blocks that way).

This is even a bit more problematic because by adding a new media query for a selector with low depth, it may change the order of the media query blocks and completely change the design of the banner, I assume that is what happened here in one of the last commits which is why this must have gone unnoticed before. I would still advocate using min-width instead of max-width since it would prevent issues of cases "between two media queries".

thank_you/templates/banner_expanded_text_var.hbs
```diff
@@ -60,12 +60,12 @@
         Gemeinsam erreichen wir mehr: Schon über 70.000 Menschen engagieren sich als Fördermitglied für eine Welt,
         in der das gesamte Wissen aller Menschen für alle frei verfügbar ist. Seien Sie dabei!
     </p>
-    <div class="btn"><a href="https://spenden.wikimedia.de/apply-for-membership?skin=1&dsn=0&type=sustaining">Ich möchte Fördermitglied werden</a></div>
+    <a href="https://spenden.wikimedia.de/apply-for-membership?skin=1&dsn=0&type=sustaining&piwik_campaign={{ CampaignName }}&piwik_kwd={{ BannerName }}"><div class="btn">Ich möchte Fördermitglied werden</div></a>
```
tzhelyazkova, line 63: why switch?

thank_you/css/styles.pcss
```diff
@@ -174,79 +186,74 @@
...
 
-				@media ( max-width: 420px ) {
-					flex-direction: column;
-				}
 				> div {
 					margin: 5px 20px;
 					flex: 1;
 					font-size: 15px;
 					.btn {
-						background: url( 'https://upload.wikimedia.org/wikipedia/commons/e/ec/Btn_white_orange-outline_1200.svg' );
```
tzhelyazkova: :disappointed:

thank_you/templates/banner_expanded_text_var.hbs
ghost: Before you had to click on the text within the button, this way you can click the complete button. :point_up:

PR #226, opened 2019-05-23
Implement survey banner for wikipedia.de

ghost:
https://phabricator.wikimedia.org/T222572

https://wikipedia.de/?banner=B19WPDE_00-190523_featurebox


tzhelyazkova (commented):
Looks good, I guess we should wait with the merging after we get the full info like the click link and so on..

PR #230, opened 2019-08-27
C19 WMDE Test 03

tzhelyazkova:
Feature: [T231210](https://phabricator.wikimedia.org/T231210)
Banners for the three skins test. 
They are identical in style and text and are based on the `var` banner of `C19 WMDE Test 01`


desktop/banner_cat17.js
```diff
@@ -1,4 +1,5 @@
 require( './css/styles.pcss' );
+require( './css/styles_var.pcss' );
```
gbirke, line 2: You should not need this in this test. All styles should be in the main style file.

desktop/css/styles_var.pcss
```diff
@@ -0,0 +1,82 @@
+:root {
```
gbirke, line 1: This should now become the new styles.pcss (or be merges with styles.pcss)

gbirke (changes requested):
I was confused when I tried to review this, because I expected 1st to set up the 3 banners with identical JS, CSS HTML from VAR (from Test 01), and then the 2nd commit would be editing the HTML (or rather passing the skin name from the banner files to the template so you can reuse the HTML).

With this PR I can't read the commits to see if the banners are identical (except for the skin name). If it's too much effort you don't need to edit the commits, but please incorporate the VAR style changes correctly.

gbirke:
After talking in person, I agree that we can use the unmodified "var" version of Test 01, since the banners won't be reused. The only thing missing is reusing the HTML template.

gbirke (approved):
After some more discussion: I'm fine with the duplicated HTML code until a change request for the banner comes around.

PR #237, opened 2019-09-23
Fix hot reloads by excluding MediaWikiTextWrapper

ghost:


gbirke (approved):
Accpted as a hot-fix, but we should rewrite the plugin for webpack 4 and check inside the plugin code.

gbirke:
A followup ticket for general improvement of the plugin is here: https://phabricator.wikimedia.org/T233596

PR #239, opened 2019-10-07
Implement Svelte-based responsive banner for Wikipedia.de

ghost:
https://phabricator.wikimedia.org/T234185


webpack.config.js
```diff
@@ -15,6 +16,40 @@ module.exports = Merge( CommonConfig, {
...
+		rules: [
+			{
+				test: /\.(html|svelte)$/,
+				exclude: /node_modules/,
+				use: {
+					loader: 'svelte-loader',
+					options: {
+						preprocess: require( 'svelte-preprocess' )( {
+							postcss: {
+								plugins: [
```
gbirke: TODO: Find a way to NOT duplicate `postcss.config.js`.

wikipedia.de_prototype/banner_ctrl.js
```diff
@@ -0,0 +1,66 @@
...
+		currentDayName: currentDayName,
+		weekdayPrepPhrase: weekdayPrepPhrase,
+		campaignDaySentence: campaignDaySentence.getSentence(),
+		campaignName: CampaignName,
+		bannerName: BannerName,
+		progressBar: progressBar.render()
+	}
+} );
+
+setTimeout( function () { progressBar.animate(); }, 1000 );
```
gbirke: This should be part of the banner initialization, using [`onMount`](https://svelte.dev/tutorial/onmount), calling a method on the `<ProgressBar>` component. Internally, the `<ProgressBar>` component can still import `'../shared/progress_bar/progress_bar'` and use that.

wikipedia.de_prototype/banner_ctrl.js
```diff
@@ -0,0 +1,66 @@
...
+const formatNumber = require( 'format-number' );
+const donorFormatter = formatNumber( { round: 0, integerSeparator: '.' } );
+
+const ProgressBar = require( '../shared/progress_bar/progress_bar' );
+
+const progressBar = new ProgressBar(
+	{ goalDonationSum: CampaignParameters.donationProjection.goalDonationSum },
+	campaignProjection,
+	{}
+);
```
gbirke: Please move this to the `<ProgressBar>` component, see also my comment below.

wikipedia.de_prototype/components/ctrl/text.svelte
```diff
@@ -0,0 +1,48 @@
...
+		<span class="text__headline--bold">Liebe Leserinnen und Leser,</span>
+		bitte verzeihen Sie die Störung. Es ist ein bisschen unangenehm, daher kommen wir gleich zur Sache.
+		{ capitalizeFirstLetter( weekdayPrepPhrase ) } {currentDayName} sind Sie in Deutschland
+		gefragt<span class="optional-text text-l">, um Wikipedias Unabhängigkeit zu sichern</span>:
+	</p>
+
+	<p class="text__paragraph">
+		{ campaignDaySentence } Wikipedia wird durch Spenden von durchschnittlich 21,48&nbsp;€ finanziert, aber 99&nbsp;% der Leserinnen und Leser spenden nicht.
+
+		<span class="text__paragraph--bold">Wenn alle, die das jetzt lesen, einen kleinen Beitrag leisten, wäre unser Spendenziel bereits am heutigen { currentDayName } erreicht.</span>
```
gbirke: We could take the chance to correct this violation of the BEM pattern that has been there since 2017. According to BEM rules,  `text__paragraph--bold` would be a modifier that gets paired with `text__paragraph` and applied to elements that have `text__paragraph`. I see this as a sub-element of the `text` block, maybe call it `text__highlight` if that doesn't clash with the highlight function in the mobile banner. Or just use `<strong>`.

wikipedia.de_prototype/variables_ctrl.pcss
```diff
@@ -0,0 +1,24 @@
...
+$banner-main-font-family: 'Arial', 'Helvetica', 'Verdana', sans-serif;
+$banner-text-font-family: 'Source Serif', 'Linux Libertine', 'Linux Libertine O', 'Georgia', 'Bitstream Charter', serif;
+
+$banner-border-radius: 0.4em;
+
+/*
+Always work 'mobile first':
+	do not use @media ( max-width: $breakpoint-x )
+	always use @media ( min-width: $breakpoint-x )
+*/
```
gbirke: 👍

webpack.common.js
```diff
@@ -71,6 +71,9 @@ module.exports = {
 		new WrapperPlugin( {
 			test: /B\d{2}WPDE.*.js$/,
 			header: function ( pageName ) {
+				if ( pageName.indexOf( 'hot-update' ) !== -1 ) {
+					return '';
+				}
```
gbirke: This change should be in a separate commit

wikipedia.de_prototype/components/ctrl/form.svelte
```diff
@@ -0,0 +1,91 @@
...
+		{ value: 'UEB', label: 'Überweisung' },
+		{ value: 'MCP', label: 'Kreditkarte' },
+		{ value: 'PPL', label: 'PayPal' },
+	];
+</script>
+
+<style lang="postcss" global>
+	@import "form.pcss";
+</style>
+
```
gbirke: Add some point we'll probably break up the form into components, but as a first step, this is already better than what we had.

wikipedia.de_prototype/banner_ctrl.js
```diff
@@ -0,0 +1,66 @@
...
+
+/* These globals are compiled into the banner through the WrapperPlugin, see webpack.common.js */
+/* global CampaignName */
+/* global BannerName */
+
+// eslint-disable-next-line no-new
```
gbirke: I've created [an issue](https://github.com/sveltejs/svelte/issues/3646) on for Svelte to see if there is a better way to appease ESLint. I've tried the [IIFE](https://medium.com/@vvkchandra/essential-javascript-mastering-immediately-invoked-function-expressions-67791338ddc6) solution by jakelucas. It works, but is confusinh to me, so I'd say we leave the `eslint-disable` here.

gbirke:
The current setup does not work on IE 11, getting the error `Function.prototype.toString: 'this' is not a Function object`

It's a "can't reproduce" closed issue in https://github.com/zloirock/core-js/issues/543 

It could be that some webpack/babel configuration magic might fix this, but for now I'm giving up.

ghost:
Ported to Preact in another PR.

PR #242, opened 2019-10-08
Fix webpack configuration for Svelte

gbirke:


gbirke:
closing, to create a new one without the wrong commit start

PR #245, opened 2019-10-10
Remove the spacing between form option buttons for certain viewports

moiikana:


wikipedia.de_prototype/ctrl/components/form.pcss
```diff
@@ -152,6 +152,27 @@
 		box-sizing: border-box;
 		display: block;
 		position: relative;
+		@media ( min-width: $breakpoint-xs ) {
+			border-right: 0;
+			border-radius: 0;
+			&:first-child {
+				border-radius: $banner-border-radius 0 0 $banner-border-radius;
+			}
+			&:last-of-type {
```
ghost, line 161: `last-child` is probably ever so slightly better because it performs better.

PR #250, opened 2019-10-21
2019 Mobile Test 01

ghost:
https://phabricator.wikimedia.org/T235728


gbirke (approved):
I'm not happy that we're using "old code" (in `banner_var.js`) from the Test02 mobile banner. This makes it hard to tell the actual difference between the 2 banners (older CSS design, different markup, missing slider) and we're missing improvements to the banner code. In the unlikely event that the VAR banner wins, and we don't pay attention, we'll lose those improvements (and now have the mental burden of having to pay attention).

I've user-tested the VAR banner and found no errors, but something *could* hide in the diff.

I approve this in for the sake of ease and speed, but it's a classic case of technical debt in the original sense.

PR #254, opened 2019-10-29
Add preact-based progress bar to desktop banner

gbirke:
Add getRemainingDays method to CampaignProjection
Add Progress Bar component for Desktop (might be turned into a mobile
one later).
Add TranslationContext for passing around translations.
Add TODOs for optimizing the Infobox component and reducing the amount
of code in banner_ctrl.js

The progress bar animation is trigged by a callback from the banner.
The progress bar animation is done via CSS transitions.

### DIFF
--- a/wikipedia.de/components/ProgressBarDesktop.jsx
+++ b/wikipedia.de/components/ProgressBarDesktop.jsx
@@ -3,6 +3,7 @@ import { Component, h } from 'preact';
 import { useContext } from 'preact/hooks';
 import formatNumber from 'format-number'
 import TranslationContext from './TranslationContext';
+import PropTypes from 'prop-types';
 import style from './ProgressBarDesktop.pcss';
 
 const NOT_STARTED = 0;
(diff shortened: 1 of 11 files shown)


wikipedia.de/components/ProgressBarDesktop.jsx
```diff
@@ -0,0 +1,71 @@
...
+			animation: NOT_STARTED
+		};
+		this.millionFormatter = formatNumber( { round: 1, prefix:'€ ', suffix: 'M', padRight: 1 } );
+		if ( props.language && this.props.language === 'DE' ) {
+			this.millionFormatter = formatNumber( { round: 1, decimal: ',', suffix: 'M €', padRight: 1 } );
+		}
+	}
+
+	componentDidMount() {
+		if ( this.props.setStartAnimation ) {
```
ghost: As discussed in person, I think it is a bit weird to have if conditions of which we know they will always be true in the code. I know this will prevent the banners from "breaking badly" but they should not make it to production in such a state anyway as the behavior would still be quietly wrong (which makes noticing and debugging it harder).

For development purposes I would see it as an advantage if they break with an error in the console because then we are more likely to notice problems.

ghost (approved):

PR #256, opened 2019-10-30
Make Matomo tracker asynchronous

gbirke:
Since the tracker library is loaded asnychronously, we need to store
the track events until it becomes available.

### DIFF
--- a/shared/matomo_tracker.js
+++ b/shared/matomo_tracker.js
@@ -1,8 +1,54 @@
+function scheduleRetryWithBackoff( tracker ) {
+	const RETRY_INTERVAL = 100;
+	setTimeout(
+		tracker.waitForTrackerToInit.bind( tracker ),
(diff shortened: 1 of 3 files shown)


ghost (approved):
Looks good to me (aside from the failing test). Just to be sure that I read it correctly, basically the code will try 10x times to find the tracker within 1 second? I probably would have made it a little more than a second just to be sure but that sounds fine with me for now.

gbirke:
I don't want the code to search for the tracker forever, so I need some kind of limit to the timeout. However, that becomes hard to test, because then the test times out or runs too long, waiting unnecessarily. I've updated the code to implement better intervals (wait for 100, 200, 300 ... ms) and pass in the re-scheduling as a function for better testability. 

This is a very good example of why code should be mostly side-effect free and how to isolate your code for side-effects. The scheduling function itself (timeout + to produce the correct sequence of millisecond intervals) is now untested, but the important part (that it stops after 10 tries) is still tested. I've user-tested it though and it works.

PR #259, opened 2019-11-04
C19 wmde en desktop test 01

moiikana:
blue border test
ticket https://phabricator.wikimedia.org/T236061

### DIFF
--- a/english/banner_ctrl.js
+++ b/english/banner_ctrl.js
@@ -269,10 +269,8 @@ function showLanguageInfoBox() {
 
 function positionLanguageInfoBox() {
 	var langInfoElement = $( '#langInfo' ),
-		formWidth = $( '#WMDE_Banner-form' ).width() - 20,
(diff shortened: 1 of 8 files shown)


english/banner_ctrl.js
```diff
@@ -132,30 +132,39 @@ function setupValidationEventHandling() {
 function appendEuroSign( field ) {
 	if ( $( field ).val() !== '' &&
 		!/^.*(€)$/.test( $( field ).val() ) ) {
-		$( field ).val( '€ ' + $( field ).val() );
+		$( field ).val( $( field ).val() + ' €' );
```
gbirke: Please keep the Euro sign in front of the number for English banners, see https://en.wikipedia.org/wiki/Euro_sign#Use

gbirke (changes requested):
Please give the elements in the CTRL also a transparent or white border to avoid the 4-pixel height difference.

Apart from that: Yay, your first banner!

gbirke (approved):

PR #260, opened 2019-11-07
Use matomo tracker for impression counting

gbirke:
Use matomo tracker for impression counting in the wikipedia.de banner.

Fix method name in tracker.

Followup to #256


gbirke:
Abandoned for #262, which also fixes the linting errors

PR #263, opened 2019-11-10
C19 WPDE 01 IE11 Fix

ghost:
Fix issues with the progress bar and the form on IE11. For larger resolutions the form is too small and on medium resolutions the form is compressed into a single row right now.


gbirke:
Thank you! 💟

PR #281, opened 2019-12-03
Set up, enable and fix linting for JSX components

ghost:
**Important:** After pulling this, you need to delete your `node_modules` directory **and** your `package.lock` and run `docker-compose run js-build npm install`, otherwise ESLint will fail on your system.


wikipedia.de/banners/Banner.jsx
```diff
@@ -1,28 +1,32 @@
-import {Component, h, Fragment } from 'preact';
+// eslint-disable-next-line no-unused-vars
+import { Component, h, Fragment } from 'preact';
+// eslint-disable-next-line no-unused-vars
```
gbirke: I think the usage of imported components in JSX should be configured in our rules with [jsx-uses-vars](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-uses-vars.md), which is already in the `eslint-plugin-react` but only needs to be configured differently.

gbirke (changes requested):

wikipedia.de/banners/Banner.jsx
ghost: I have tried that but whenever I add `"jsx-uses-vars": 1` to the rules, ESLint seems to bug out and stops recognizing ` eslint-disable-next-line no-unused-vars` altogether even though we still need it. If you'd like, you can double-check and see if you can find a solution but I am bit out of ideas at this point 🤔 .

gbirke (approved):

PR #284, opened 2019-12-09
C19 wmde test 17

moiikana:
https://phabricator.wikimedia.org/T240202


gbirke (approved):
Interesting idea to leave the text template as-is and use CSS instead. Totally works and saves an additional template. For bigger text changes, we'll have to structure  the text markup differently to see with one glance which is which, but for this small text change it's perfect.

PR #286, opened 2019-12-11
C19 wmde en desktop test 03

moiikana:
https://phabricator.wikimedia.org/T240241

### DIFF
--- a/english/templates/banner_text.hbs
+++ b/english/templates/banner_text.hbs
@@ -17,5 +17,9 @@
     Wikipedia is a place to learn, not a place for advertising.
     It unites all of us who love knowledge: contributors, readers and the donors who keep us thriving.
     The heart and soul of Wikipedia is a community of people working to bring you unlimited access to reliable, neutral information.
-    Please take a minute to help us keep Wikipedia growing. Thank you!
+    <span class="ctrl-text">Please take a minute to help us keep Wikipedia growing.</span>
+    <span class="var-text">If Wikipedia gave you <span class="no-wrap">€ 5</span> worth of knowledge this year, take a minute to donate and help us to keep Wikipedia growing.</span>
+    Thank you!
+    <span class="ctrl-text--invisible">______________I'm an invisible placeholder____________________</span>
+
 </p>
--- a/campaign_info.toml
+++ b/campaign_info.toml
@@ -106,19 +106,19 @@ tracking = "wpde-02-191129-var"
 
 # English Campaign
 [english]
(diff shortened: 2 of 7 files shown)


english/templates/banner_text.hbs
```diff
@@ -17,5 +17,9 @@
     Wikipedia is a place to learn, not a place for advertising.
     It unites all of us who love knowledge: contributors, readers and the donors who keep us thriving.
     The heart and soul of Wikipedia is a community of people working to bring you unlimited access to reliable, neutral information.
-    Please take a minute to help us keep Wikipedia growing. Thank you!
+    <span class="ctrl-text">Please take a minute to help us keep Wikipedia growing.</span>
+    <span class="var-text">If Wikipedia gave you <span class="no-wrap">€ 5</span> worth of knowledge this year, take a minute to donate and help us to keep Wikipedia growing.</span>
+    Thank you!
+    <span class="ctrl-text--invisible">______________I'm an invisible placeholder____________________</span>
```
ghost, line 23: It's better to re-use the exact same text again to be sure the banner will break at the same time. Usually we create a banner_text_var for this but if you want to stick to one _text.hbs that's fine. In that case it would look like that:

```
<span class="ctrl-text--invisible">If Wikipedia gave you <span class="no-wrap">€ 5</span> worth of knowledge this year, take a minute to donate and help us to keep Wikipedia growing.</span>
<span class="var-text--invisible">Please take a minute to help us keep Wikipedia growing.</span>
```

And then ctrl-text-invisible would be "hidden" on VAR and var-text--invisible would be 'display:none' on VAR (and vice versa on CTRL).

moiikana: maybe I thought too complicated about that again, but I needed to keep the banners the exact same height and tried to achieve it with this invisible text. The --invisible span contains the exact amount of characters that is needed to have the same total amount of characters in CTRL and VAR (it felt very horrible doing this but it seemed to be the quickest and most effective way to achieve equal height here without having to add lots of extra space for both banners and making them unnecessarily big).
_length(ctrl-text--invisible) = length(var-text) - length(ctrl-text)_

ghost: Hey, the problem with that is that most fonts are not "monospaced" which means different characters have different lengths, especially across browsers / operating systems. "W" and "_" are generally wider characters than "i" or "t". That's why the only way to really get it right is to make an exact copy of the text. I personally don't care that much but Till and the gang had complaints about this in the past.

PR #289, opened 2019-12-16
Implement use of funds modal

ghost:
https://phabricator.wikimedia.org/T239622


shared/components/FundsModal.jsx
```diff
@@ -0,0 +1,86 @@
+// eslint-disable-next-line no-unused-vars
+import { Component, h } from 'preact';
+import classNames from 'classnames';
+
+export default class FundsModal extends Component {
+	constructor( props ) {
+		super( props );
+		this.locale = props.locale;
+	}
+	formatAmount = function ( amount ) {
```
gbirke, line 10: Maybe pass in numberFormatters (initialized from the `format-number` library) as properties and only do the-space-to-number conversion in the component.

shared/components/FundsModal.jsx
```diff
@@ -0,0 +1,86 @@
...
+		if ( targetElement.classList.contains( 'is-active' ) ) {
+			targetElement.classList.remove( 'is-active' );
+		} else {
+			let accordionElements = Array.from( document.querySelectorAll( '.banner_modal_accordion' ) );
+			for ( let element of accordionElements ) {
+				element.classList.remove( 'is-active' );
+			}
+			targetElement.classList.add( 'is-active' );
+		}
+	};
```
gbirke, line 33: Ok for now, but to my knowledge it's not a good practice to manipulate the DOM that was rendered by React. Since we're using React as a templating engine, it's ok, but future versions of this component should use reactive sub-components.

gbirke (approved):

PR #297, opened 2020-01-07
Implement desktop banners with preact

gbirke:


shared/banner_presenter.js
```diff
@@ -0,0 +1,96 @@
...
+		this.sizeTrackRatio = sizeTrackRatio;
+		this.appearanceDelay = appearanceDelay;
+	}
+
+	present( Banner, bannerContainer, props ) {
+		const skinAdjuster = getSkinAdjuster();
+
+		const sizeIssueIndicator = new SizeIssueIndicator( skinAdjuster.getSizeIssueThreshold() );
+
+		skinAdjuster.moveBannerContainerToTopOfDom();
```
gbirke: We should move this down, after the check if the banner should be presented

shared/banner_presenter.js
```diff
@@ -0,0 +1,96 @@
...
+
+		if ( onMediaWiki() && !mediaWikiIsShowingContentPage() ) {
+			return;
```
gbirke, line 27: We should call `mw.centralNotice.setBannerLoadedButHidden();` before returning.

PR #301, opened 2020-01-14
Implement german Ipad banner with preact

moiikana:


pad_preact/components/ui/form/DonationForm.jsx
```diff
@@ -0,0 +1,147 @@
...
+			setdisabledGiveAddressOptions( [] );
+		}
+	};
+
+	const formAction = addressOption === GiveAddressOptions.YES.value ?
+		'https://spenden.wikimedia.de/donation/new?piwik_campaign=' : 'https://spenden.wikimedia.de/donation/add?piwik_campaign=';
+
+	return <div className="form">
+		<form method="post" name="donationForm" className="form__element"
+			action={ formAction }>
```
gbirke: The campaign parameters (`props.bannerName` and `props.campaignName`) are missing.

pad_preact/components/ui/form/FormItemsBuilder.js
```diff
@@ -0,0 +1,59 @@
...
+	PAYPAL: { value: 'PPL', label: 'payment-paypal' },
+	SOFORT: { value: 'SUB', label: 'payment-sofort' }
+};
+
+export const GiveAddressOptions = {
+	YES: { value: 'person', label: 'give-address-statement-positive' },
+	NO: { value: 'anonym', label: 'give-address-statement-negative' }
+};
+
+export default class FormItemsBuilder {
```
gbirke: Since the other form items are 100% identical, please use inheritance here. This will make the class much shorter.

shared/messages/de.js
```diff
@@ -48,7 +48,10 @@ const Translations = {
 	'payment-bank-transfer': 'Überweisung',
 	'payment-credit-card': 'Kreditkarte',
 	'payment-paypal': 'PayPal',
-	'payment-sofort': 'Sofortüberweisung'
+	'payment-sofort': 'Sofortüberweisung',
+
+	'give-address-statement-positive': 'Ja',
+	'give-address-statement-negative': 'Nein'
```
gbirke, line 55: The translations file is part of every banner, not only the iPad banner. Therefore, we should keep a certain discipline to mark translations that are part of an A/B test as such, to potentially clean up after the tets is done.

gbirke (changes requested):
Two small changes are needed.

I'm torn on the `ui` and `ui/form` subfolder in `components`. It keeps consistency, but at the cost of deeply nested and confusing relative paths. What's more important to you?

pad_preact/Banner.jsx
```diff
@@ -0,0 +1,117 @@
...
+				</TranslationContext.Provider>
+			</BannerTransition>
+			<FundsModal
+				fundsModalData={props.fundsModalData}
+				toggleFundsModal={ this.toggleFundsModal }
+				isFundsModalVisible={ this.state.isFundsModalVisible }
+				locale='de'/>
+		</div>;
+	}
+
```
gbirke, line 125: Please remove the `FundModal` component and all state and state-changing methods from this banner. For iPad, the banner just displays a link.

pad_preact/Banner.jsx
```diff
@@ -0,0 +1,117 @@
...
+const CLOSED = 2;
+
+export default class Banner extends Component {
+
+	static propTypes = {
+		/** callback when banner closes */
+		onClose: PropTypes.func,
+		/** */
+		registerDisplayBanner: PropTypes.func.isRequired
+	}
```
gbirke, line 43: I'm not sure what's worse - an incomplete list of available properties or no list of properties at all. What do you think?

pad_preact/Banner_var.jsx
```diff
@@ -0,0 +1,117 @@
...
+			</BannerTransition>
+			<FundsModal
+				fundsModalData={props.fundsModalData}
+				toggleFundsModal={ this.toggleFundsModal }
+				isFundsModalVisible={ this.state.isFundsModalVisible }
+				locale='de'/>
+		</div>;
+	}
+
+}
```
gbirke: Please delete `Banner_var.jsx`, that's unneccessary duplication. Instead, load different form components and pass in the form component as a property (like we do for BannerText)

shared/messages/de.js
```diff
@@ -48,7 +48,11 @@ const Translations = {
 	'payment-bank-transfer': 'Überweisung',
 	'payment-credit-card': 'Kreditkarte',
 	'payment-paypal': 'PayPal',
-	'payment-sofort': 'Sofortüberweisung'
+	'payment-sofort': 'Sofortüberweisung',
+
+	// the following 2 lines are only for the ipad var test, remove if not in use anymore
```
gbirke, line 53: 👍

pad_preact/styles/styles_ctrl.pcss
```diff
@@ -0,0 +1,12 @@
...
+@import '../../shared/components/BannerTransition.pcss';
+
+@import './Banner.pcss';
+@import './DonationForm.pcss';
+@import './SelectGroup.pcss';
+@import './Footer.pcss';
+@import './Infobox.pcss';
+@import './BannerText.pcss';
+@import './ProgressBar.pcss';
+@import '../../shared/components/ui/FundsModal.pcss';
```
gbirke: You can remove this line, as we don't have a FundsModal here

gbirke (changes requested):

pad_preact/Banner.jsx
moiikana: would a complete list of properties in this case be the following? (+their respective types) : 
- trackingIds: ,
- campaignParameters: ,
- campaignProjection: ,
- formatters: ,
- bannerText: ,
- donationForm: ,
- translations: ,
- formItems:
- trackingData: this.trackingData,
- impressionCounts:,
- onClose: PropTypes.func,
- registerDisplayBanner: PropTypes.func.isRequired

these are the properties from banner_ctrl.js, banner_presenter.js and Banner.jsx.
I think it would be nice to have them documented like this, if this list is complete and correct.

pad_preact/Banner.jsx
```diff
@@ -0,0 +1,118 @@
...
+import { CampaignProjection } from '../shared/campaign_projection';
+
+const PENDING = 0;
+const VISIBLE = 1;
+const CLOSED = 2;
+
+export default class Banner extends Component {
+
+	static propTypes = {
+		campaignParameters: PropTypes.object,
```
moiikana, line 23: I tried to accumulate the propTypes that are passed down to the Banner class by banner_ctrl.js, banner_presenter and so on..
Is it correct that they all should be defined here? (they are used here, so maybe yes..)
I tried to give them the most fitting type and tried to check my guessings with `console.log(typeof ....)` , but that always told me it's either an object or a func. I'm also not sure about whether it could be PropTypes.elementType or .element in line 23-24.

pad_preact/Banner.jsx
```diff
@@ -0,0 +1,118 @@
...
+		campaignProjection: PropTypes.instanceOf( CampaignProjection ), //object?
+		formatters: PropTypes.object,
+		bannerText: PropTypes.elementType, //func?
+		donationForm: PropTypes.elementType, //func?
```
gbirke: `elementType` is exactly right. Elements can be either functions or classes.

pad_preact/Banner.jsx
```diff
@@ -0,0 +1,118 @@
...
+		translations: PropTypes.object,
+		formItems: PropTypes.object,
```
gbirke, line 33:
```suggestion
                /**
                   *  Contains items with values and labels for SelectGroup components 
                   *  in the form, e.g. amounts, payment types, intervals, etc 
                   */
		formItems: PropTypes.object,
```

pad_preact/Banner.jsx
```diff
@@ -0,0 +1,118 @@
...
+
+		/** complex object that holds any data needed for tracking purposes */
+		trackingData: PropTypes.object,
+		impressionCounts: PropTypes.instanceOf( LocalImpressionCount ), //object?
+
+		/** callback when banner closes */
+		onClose: PropTypes.func,
+		registerDisplayBanner: PropTypes.func.isRequired
```
gbirke, line 42:
```suggestion
                /** Callback to register a displayBanner function with the BannerPresenter */
		registerDisplayBanner: PropTypes.func.isRequired
```

pad_preact/Banner.jsx
```diff
@@ -0,0 +1,118 @@
...
+	static propTypes = {
+		campaignParameters: PropTypes.object,
+		campaignProjection: PropTypes.instanceOf( CampaignProjection ), //object?
+		formatters: PropTypes.object,
+		bannerText: PropTypes.elementType, //func?
+		donationForm: PropTypes.elementType, //func?
+		translations: PropTypes.object,
+		formItems: PropTypes.object,
+
+		/** complex object that holds any data needed for tracking purposes */
```
gbirke:
```suggestion
		/** object that holds any data needed for tracking purposes */
```

pad_preact/Banner.jsx
gbirke: Thank you so much! 

Two props are missing: `bannerName` and `campaignName` (coming from `...trackingIds`) in the entry point.

PR #303, opened 2020-01-20
Implement english ipad banners with preact

moiikana:
https://phabricator.wikimedia.org/T243208


pad_english_preact/components/ui/Footer.jsx
```diff
@@ -0,0 +1,40 @@
...
+						<span className="footer__importantinfo"><span
+							className="iban-block">
+							<span>DE33</span><span>1002</span><span>0500</span><span>0001</span><span>1947</span><span>00</span></span></span>
+					</div>
+				</div>
+			</div>
+
+			<div className="footer__item footer__item--rightflex">
+				<a id="application-of-funds-link" className="application-of-funds-link"
+					href="https://spenden.wikimedia.de/use-of-funds?skin=0&piwik_campaign={{ CampaignName }}&piwik_kwd={{ BannerName }}_link">
```
gbirke: You need to build the proper link URL (`campaignName` and `bannerName` must come from props).

pad_english_preact/components/ui/Footer.jsx
```diff
@@ -0,0 +1,40 @@
...
+					{ Translations[ 'use-of-funds-link' ] }
+				</a>
+			</div>
+
+		</div>
+	</div>;
+}
```
gbirke, line 40: The difference between the two footers is the "Application of Funds" behavior - either showing a modal box or going to a new page. To reduce duplication I'd prefer one shared footer with a `showFundsModal` prop. We either pass an event handler into `showFundsModal` (that toggles the modal and calls `e.preventDefault()` on the original event) or an empty function. The empty function won't do anything and browser will execute the original behavior of the link - opening a new page.

pad_english_preact/components/BannerText_var.jsx
```diff
@@ -0,0 +1,38 @@
+// eslint-disable-next-line no-unused-vars
+import { h } from 'preact';
+
+// eslint-disable-next-line no-unused-vars
+function capitalizeFirstLetter( message ) {
+	return message.charAt( 0 ).toUpperCase() + message.slice( 1 );
+}
```
gbirke: This function is not needed in these BannerText instances, please remove it from both.

shared/components/ui/LanguageWarningBox.jsx
```diff
@@ -0,0 +1,13 @@
...
+import classNames from 'classnames';
+import TranslationContext from '../TranslationContext';
+
+export default function LanguageWarningBox( props ) {
+	const Translations = useContext( TranslationContext );
+
+	return <div className={ classNames( props.show ? 'lang-warning--visible' : 'lang-warning--hidden', 'lang-warning' ) }>
+		<p>{ Translations[ 'language-warning' ] }</p>
+	</div>;
+}
```
gbirke, line 10: Since the `LanguageWarningBox` is only shown in English banners, we could gain some readability and save a few bytes and CPU cycles by removing `TranslationContext ` and putting the string directly in here.

gbirke (changes requested):
Thank you for that expressive commit message, our future selves will be very happy about that.

pad_english_preact/components/ui/Footer.jsx
```diff
@@ -0,0 +1,40 @@
...
+
+export default function Footer() {
+	const Translations = useContext( TranslationContext );
+
+	return <div className="banner__footer">
+		<div className="footer">
+
+			<div className="bank-info">
+				<div className="footer__item">
+					<span>Spendenkonto Wikimedia Foerdergesellschaft</span>
```
gbirke: In the new shared footer the word "Spendendkonto" must be translatable, it's 'DONATION ACCOUNT' ( poabbly uppercased with CSS) in the old english iPad banner

pad_english_preact/components/ui/Footer.jsx
moiikana: when will I ever notice this beforehand ...🤦‍♂️ thx

gbirke (approved):

PR #311, opened 2020-03-06
Unify custom amount field

Abban:
https://phabricator.wikimedia.org/T244714

### DIFF
--- a/shared/components/ui/form/SelectCustomAmount.jsx
+++ b/shared/components/ui/form/SelectCustomAmount.jsx
@@ -1,10 +1,8 @@
-import { Component, createRef } from 'preact/compat';
 // eslint-disable-next-line no-unused-vars
-import { h } from 'preact';
+import { h, Component } from 'preact';
 import classNames from 'classnames';
 
 export default class SelectCustomAmount extends Component {
-	radioInput = createRef();
 
 	constructor( props ) {
 		super( props );
@@ -16,17 +14,9 @@ export default class SelectCustomAmount extends Component {
 
 	onFocus = ( e ) => {
 		this.setState( { showEuro: true, focused: true } );
-		this.radioInput.current.click();
 		if ( this.props.value !== '' ) {
 			e.target.select();
 		}
-	};
-
-	onInput = ( e ) => {
-		if ( !this.radioInput.current.checked ) {
-			this.radioInput.current.click();
-		}
-		this.props.onInput( e );
 	};
 
 	onBlur = ( e ) => {
@@ -39,7 +29,7 @@ export default class SelectCustomAmount extends Component {
 
 	render( props, state ) {
 		return <label className="select-group__option select-group__option--amount-other-input">
-			<input type="radio" name={ props.fieldname } className="select-group__input" value="" ref={ this.radioInput }/>
+			<input type="radio" name={ props.fieldname } className="select-group__input" value="" checked={ state.focused || props.value !== null }/>
 
 			<div className={ classNames(
 				'select-group__custom-input',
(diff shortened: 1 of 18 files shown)


moiikana (commented):
I approve this but would like to have @gbirke 's opinion on the SelectCustomAmount.jsx class, about whether there might have been more preact-optimized ways to handle the user events or if this is totally fine.

shared/number_formatter/de.js
```diff
@@ -3,6 +3,7 @@ import amountForServerFormatter from './server';
 
 export const integerFormatter = formatter( { round: 0, integerSeparator: '.' } );
 export const amountInputFormatter = formatter( { round: 2, suffix: ' €', decimal: ',', integerSeparator: '' } );
+export const customAmountInputFormatter = formatter( { round: 2, decimal: ',', integerSeparator: '' } );
```
gbirke, line 7: When we drop the old-style banners we can also drop the `amountInputFormatter`. Please add a `/** @deprecated */` comment to the `amountInputFormatter` to remind us and avoid future confusion over which one to use.

shared/components/ui/form/SelectCustomAmount.jsx
```diff
@@ -1,18 +1,68 @@
...
+		super( props );
+		this.state = {
+			focused: false,
+			showEuro: false
+		};
+	}
+
+	onFocus = ( e ) => {
+		this.setState( { showEuro: true, focused: true } );
+		this.radioInput.current.click();
```
gbirke: Have you tried setting the `checked` attribute of the input tag based on the current state? References and native DOM elements should be a last resort.

shared/components/ui/form/SelectCustomAmount.jsx
```diff
@@ -1,18 +1,68 @@
...
+					maxLength="8"
+					onFocus={ this.onFocus }
+					onBlur={ this.onBlur }
+					autoComplete="off"
+					placeholder={ props.placeholder }
+					className="select-group__custom-input--input" />
+			</div>
+		</label>;
+	}
+
```
gbirke, line 57: I'm torn on the implementation. I'd have liked it more if `SelectCustomAmount` had stayed stateless and the blur and value changes and the euro sign would be implemented purely with CSS (see https://spenden.wikimedia.de/ ). The performance would be better and it would be less code. OTOH the JavaScript makes the behavior (or at least the intended styling "hooks" (i.e. CSS classes) very explicit and more readable.

This is not a "please change" recommendation, more like a reminder to use CSS instead of JS if possible.

gbirke (changes requested):

shared/components/ui/form/SelectCustomAmount.jsx
Abban: Yes, I found that it had a weird behaviour where it added the attribute onto the element in the DOM and prevented the user from being able to select the other radio options.

shared/number_formatter/de.js
Abban: Is there a preferred style for this? In front of the variable, like a docblock or trailing at the end of the line like in CSS?

shared/components/ui/form/SelectCustomAmount.jsx
Abban: Yeah this was a tricky one, some of the required behaviour of the euro symbol depends on js (keeping it visible when the field has a value) so I decided to keep the code that controls that together rather than split it between js/css.

shared/number_formatter/de.js
gbirke: Like a docblock

gbirke (approved):

PR #312, opened 2020-03-09
Re-Refactor banners for wikipedia.de

moiikana:
https://phabricator.wikimedia.org/T243209

- Reorganize classes and files to use shared components.
- in order to reduce the complexity and size of this banner, for this PR only the desktop banner is used on wikipedia.de, the mobile ones will not be shown, but the CSS for them is still archived in their folders.


gbirke:
I found out why the slides are not displaying on mobile - the Desktop banner display code fails before the mini banner initialization code is run. The culprit here is the line `setStartAnimation={this.registerStartProgressbar}` in `DesktopBanner.jsx`. The `registerStartProgressbar` method does not exist yet in the wikipedia.de version of the Dektop banner, plz compare to the Desktop wp.org banner and add code accordingly (`registerStartProgressbar`, call `this.startprogressBar` when the Dektip banner has finished animating).

How i found this: Looking at the webdev console in the browser showed an error where the property registerStartProgressbar was empty. Searching for occurrences in the code led me to DektopBanner.jsx. The console is my go-to place whenever an error occurs. many times you'll find an error message there .

gbirke (approved):

Abban (approved):

PR #314, opened 2020-03-13
Fix mobile banners

gbirke:
Safari mobile crashes on close when the Minerva SkinAdjuster moves the
banner element to the top of the DOM.

This change adapts the Transition elements to be sgtill absoluetly
positioned, but inside a relatively positioned parent element,
whose "top" position is changed by skinAdjuster. To keep the 2-phase
animation, the FollowupTransition must have a negative top offset that
corresponds to the positive top offset of its parent element. That way,
it will always stay at position 0. It also needs to incorporate the
previous offset of the mini banner.

Renamed "frbanner" to "fullpage-banner" to properly show and hide the 2
banners depending on state and to have more semantic names.

### DIFF
--- a/shared/skin/minerva.js
+++ b/shared/skin/minerva.js
@@ -39,10 +39,6 @@ export default class Minerva extends Skin {
 		$( '#ca-ve-edit, .mw-editsection-visualeditor' ).click( onEdit );
 	}
 
-	moveBannerContainerToTopOfDom() {
-		$( 'body' ).prepend( $( '#centralNotice' ) );
-	}
-
 	getSizeIssueThreshold() {
 		return 180;
 	}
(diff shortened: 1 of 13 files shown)


shared/skin/minerva.js
```diff
@@ -40,7 +40,7 @@ export default class Minerva extends Skin {
 	}
 
 	moveBannerContainerToTopOfDom() {
-		$( 'body' ).prepend( $( '#centralNotice' ) );
+		// $( 'body' ).prepend( $( '#centralNotice' ) );
```
Abban: Is this supposed to be commented?

gbirke: Nah, that was just part of trying stuff out, the whole function should be deleted from this implementation (to use the default empty implementation of the abstract parent class).

Abban (approved):
Approving and leaving for @moiikana to review before merging as these changes may affect the refactor ticket

moiikana (approved):

PR #316, opened 2020-03-18
Fix for mobile banner space miscalculation

Abban:
On orientation change this does the following:

- When mini banner is displayed, forces the slider to update size before calculating banner size
- When full page banner is displayed, forces it to re-render

https://phabricator.wikimedia.org/T247226


gbirke (commented):
As discussed, please add a debounce of 200ms to the resize handler

PR #329, opened 2020-07-01
Investigate Event Tracking

Abban:
https://phabricator.wikimedia.org/T243625


shared/event_logging_tracker.js
```diff
@@ -1,82 +1,51 @@
...
-	trackCloseEventViewPortDimensions( $trackedElement, dimensionCallback, slidesShown, finalSlide, trackingRatio = 0.01, bannerEvent = 'banner-closed' ) {
-		$trackedElement.click( function () {
-			if ( this.trackBannerEvent( bannerEvent, slidesShown, finalSlide, trackingRatio ) ) {
-				// also track viewport dimensions on top of recording the banner closed event.
-				this.trackViewPortDimensionsOnEvent( dimensionCallback(), VIEWPORT_TRACKING_CLOSED_EVENT_IDENTIFIER, 1 );
-			}
-		}.bind( this ) );
+	trackBannerEventWithViewport( actionName, slidesShown, finalSlide, trackingRatio = 0.01, dimensionData = {} ) {
+		if ( this.trackBannerEvent( actionName, slidesShown, finalSlide, trackingRatio ) ) {
+			this.trackViewportData( VIEWPORT_TRACKING_CLOSED_EVENT_IDENTIFIER, dimensionData, 1 );
```
gbirke: You could make this `1` an exported constant like `ALWAYS_TRACK` or `TRACK_100_PERCENT` (maybe you can find a better name).

test/event_logging_tracker_test.js
```diff
@@ -0,0 +1,152 @@
...
+
+	get() {
+		return this.randomNumbers[ this.currentIndex++ ];
+	}
+}
+
+describe( 'EventLoggingTracker', function () {
+
+	it( 'contains all required methods', () => {
+		const tracker = new EventLoggingTracker( BANNER_NAME, { track: sinon.spy() }, () => 0 );
```
gbirke: The readability of `() => 0` could be improved by defining it like 

    const fixedRandom = () => 0

and re-using the function name. could also be called `dummyRandom` or `zeroRandom` or smth.

test/event_logging_tracker_test.js
```diff
@@ -0,0 +1,152 @@
...
+		const eventSpy = { track: sinon.spy() };
+		const tracker = new EventLoggingTracker( BANNER_NAME, eventSpy, () => 0 );
+
+		tracker.trackSizeIssueEvent( DIMENSION_DATA, eventData.eventRate );
+
+		assert.deepEqual( eventSpy.track.firstCall.args[ 0 ], BANNER_SIZE_SCHEMA );
+		assert.deepEqual( eventSpy.track.firstCall.args[ 1 ], eventData );
+	} );
+
+	it( 'Only sends when rate is higher than random', () => {
```
gbirke, line 88: I find this test a bit confusing, esp. with the maths done to `randomNumbers`. What is it that you're testing? Do you use multiple random numbers to build a certain sample size (just to be sure)?

Could you find a way to test tuples of `[randomNumber, ratio]`? That way, when reading the code the reader doesn't have to jump between the random number array and the call to `trackBannerEvent` and doesn't have to do maths in their head

gbirke (changes requested):
Requesting changes because I found the test too confusing 😓

shared/event_logging_tracker.js
```diff
@@ -1,82 +1,53 @@
...
 const VIEWPORT_TRACKING_CLOSED_EVENT_IDENTIFIER = 'vtc';
 
-export default class EventLoggingTracker {
+export const EVENT_SCHEMA = 'event.WMDEBannerEvents';
+export const BANNER_SIZE_SCHEMA = 'event.WMDEBannerSizeIssue';
 
-	constructor( bannerName ) {
-		this.bannerName = bannerName;
-	}
+const ALWAYS_TRACK = 1;
```
gbirke:
```suggestion
export const ALWAYS_TRACK = 1;
```

This can then also be used in banner code, making it more readable

test/event_logging_tracker_test.js
```diff
@@ -0,0 +1,106 @@
...
+
+const BANNER_NAME = 'TestTracker';
+const DIMENSION_DATA = {
+	window: {
+		width: 10,
+		height: 10
+	},
+	bannerHeight: 10
+};
+const ALWAYS_TRACK = 1;
```
gbirke: You can `import` the one from the tracker here.

gbirke (approved):

PR #330, opened 2020-07-07
[DNM] Collage Banner

Abban:
https://phabricator.wikimedia.org/T256609


gbirke (commented):
No objections to the code changes so far. 

Is the fixed height for the .banner-text-copy necessary, or are you just using it as a stopgap measure until the form determines the height? Usually we don't set heights in layout elements to better support responsiveness

desktop_preact/Banner.jsx
```diff
@@ -80,18 +82,24 @@ export default class Banner extends Component {
...
 	render( props, state, context ) {
 		const campaignProjection = props.campaignProjection;
+
 		return <div
 			className={classNames(
 				'wmde-banner',
 				state.displayState === CLOSED ? 'wmde-banner--hidden' : '',
-				state.displayState === VISIBLE ? 'wmde-banner--visible' : ''
+				state.displayState === VISIBLE ? 'wmde-banner--visible' : '',
+				state.isAmountToggleOpen ? 'wmde-banner--amount-toggle-open' : '',
```
gbirke: FYI, you can also provide objects with boolean values to toggle class names, e.g.
```javascript
{
   'wmde-banner--hidden': state.displayState === CLOSED, 
   'wmde-banner--visible': state.displayState === VISIBLE,
   'wmde-banner--amount-toggle-open': state.isAmountToggleOpen
}
```

This can help with the readability when all the classes come first and the conditions second and you don't have to provide empty strings in ternary operators. But ultimately it's a matter of opinion and taste and we mix and match the styles thoughout the code base.

shared/components/ui/form/DonationForm.jsx
```diff
@@ -50,6 +52,7 @@ export default function DonationForm( props ) {
 					currentValue={ paymentInterval }
 					onSelected={ e => setInterval( e.target.value ) }
 					disabledOptions={ disabledIntervals }
+					errorPosition={ props.errorPosition }
```
gbirke, line 59: If this banner leaves the prototype stage we should definitvely document all the props (Like we do in `ProgressBar.js`). This is becoming confusing and I wouldn't know what value goes in here and where it comes from.

shared/components/ui/form/SelectGroup.jsx
```diff
@@ -2,7 +2,18 @@
 import { h } from 'preact';
 import classNames from 'classnames';
 
-export default function SelectGroup( props ) {
+export const ERROR_POSITION_TOP = 1;
+export const ERROR_POSITION_BOTTOM = 2;
```
gbirke: Good enough, but with potential for errors when we define the prop as `number`. Since this is an enum, we could potentialle use one of the following ES6 kludges to define it a bit more semantic/type-safe: https://stackoverflow.com/questions/44447847/enums-in-javascript-with-es6

gbirke (approved):
Just some suggestions for polishing the code, the requested change is in #333

PR #335, opened 2020-07-15
Fix Banner Space Errors

Abban:
Caused by a bug discovered here: https://phabricator.wikimedia.org/T256611#6304465

### DIFF
--- a/desktop_preact/Banner.jsx
+++ b/desktop_preact/Banner.jsx
@@ -63,6 +63,12 @@ export class Banner extends Component {
 		}
 	}
 
+	onFinishedTransitioning = () => {
+		this.props.onFinishedTransitioning();
+		// Uncomment when we have a progress bar during the campaign again
+		// this.startProgressbar()
+	}
+
 	closeBanner = e => {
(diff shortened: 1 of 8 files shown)


desktop_preact/Banner.jsx
```diff
@@ -63,6 +63,10 @@ export class Banner extends Component {
 		}
 	}
 
+	onFinishedTransitioning = () => {
+		this.props.onFinishedTransitioning();
```
gbirke, line 67: This is just a suggestion, that goes against the "You should not use comments as version control" dogma, but might help our future selves to determine where to put the progress bar start. But maybe we're clever enough to figure that out ...

```suggestion
		this.props.onFinishedTransitioning();
		// Uncomment when we have a progress bar during the campaign again
		// this.startProgressbar()
```

gbirke (approved):

desktop_preact/Banner.jsx
Abban: I also added the register progress bar callback method back in

PR #336, opened 2020-07-15
Remove Old Banners

Abban:
https://phabricator.wikimedia.org/T257790


Abban:
Can I remove any of the other folders?

moiikana:
wikipedia.de_OLD can be removed as well, it was just an old clone to make development easier

PR #341, opened 2020-07-22
Reintroduce logging the viewport dimensions on submit

moiikana:
https://phabricator.wikimedia.org/T258266


Abban:
We should also add a new test in `event_logging_tracker_test.js` that makes sure that the `vtc` prefix is appended to the banner name when `trackBannerEventWithViewport` happens.

PR #342, opened 2020-07-23
Cleanup for de desktop banners

Abban:
This fixes margin collapsing errors and adds some better visual consistency to the layout.

I also added a temporary new footer with input boxes that allow auto selection of the bank details for copy/paste purposes. If accepted this should be moved into shared and the other banner styles updated to accomodate.

I also also added pure css radio replacements to help with vertical alignment across browsers


Abban:
Tested in IE11 and Edge Legacy

PR #345, opened 2020-08-04
C20_WMDE_Test_03 Desktop Banner

moiikana:
Precampaign desktop DE test 03

- based off C20_WMDE_Test_01 design (no 2nd page)

https://phabricator.wikimedia.org/T259463


gbirke (changes requested):
Code looks good, thanks for cleaning up!

Please use the new style of rendering banner text to ensure same banner height: Rendering CTRL and VAR Text side-by-side in both banners, in a container with overflow-clipping, using CSS classes to move the CTRL and VAR element into view (see 1st desktop test for an example). 

Technically, the 2nd commit should only contain the addition of the `desktop/components/BannerText_var.jsx` file, its integration into the banner and the additional CSS. You have some changes to CTRL there, that would belong in the 1st commit, if we adhere to the dogma. But if it'S too much trouble, you can amend the commit.

gbirke (approved):

PR #346, opened 2020-08-05
C20 wmde mobile test 01

Abban:
https://phabricator.wikimedia.org/T259461


moiikana (commented):
in my desktop test 03 PR I moved the new footer to shared...EasySelectFooter.jsx
maybe you could do that here, too

shared/banner_slider.js
```diff
@@ -48,6 +48,9 @@ export class Slider {
 	 * @param {int} milliseconds - Time to wait before the slider starts
 	 */
 	enableAutoplayAfter( milliseconds = 0 ) {
+		if ( !this.sliderAutoPlaySpeed ) {
```
gbirke: The doc comment for the constructor says that it's an `int`, so this was immediately confusing to me. I'd rather not repeat the laissez-faire typing done by Flickity. 

I can think of two options to change this:

* change the constructor docblock to document the type to `{int|false}` (don't know if that's even a valid type) and explain what `false` means or
*  explain the special value `0` in the constructor docblock and cast to boolean in `initialize`, depending on the param being 0 or > 0

Choose whatever feels less wrong to you.

mobile/banner_var.js
```diff
@@ -24,7 +24,7 @@ const trackingIds = getTrackingIds( bannerContainer );
 const trackingData = createTrackingData( trackingIds.bannerName );
 const bannerPresenter = new BannerPresenter(
 	trackingData,
-	bannerContainer.dataset.delay || 5000,
+	bannerContainer.dataset.delay || 7500,
```
gbirke: Plz ask our PM if 7.5 seconds is standard across banners. AFAIR the default dealy was always 5 seconds for mobile.

gbirke (changes requested):

shared/banner_slider.js
Abban: Thinking about it the Slider object probably shouldn't care about the autoplay speed there anyway. I'll add a prop higher up the tree.

PR #347, opened 2020-08-06
C20 wmde desktop test 04

moiikana:
text change only

https://phabricator.wikimedia.org/T259466


desktop/components/BannerText.jsx
```diff
@@ -42,8 +40,7 @@ export default function BannerText( props ) {
 					<span className="optional-text text-m">Wenn Wikipedia eine kommerzielle Seite sein würde, wäre das ein riesiger Verlust für die Welt. </span>
 					Sicher könnten wir mit Werbung eine Menge Geld verdienen. Aber dann wäre Wikipedia komplett anders.
 					Wir könnten ihr nicht vertrauen.{ ' ' }
```
Abban: You only need to add these `{ ' ' }`  when the element on the next line is an opening tag. The parser leaves a single space between plain text with linebreaks

Abban (approved):

PR #348, opened 2020-09-03
C20 wmde desktop test 05

moiikana:
Introduces info header in the form section and adds a info icon in the banner text.
https://phabricator.wikimedia.org/T261620


Abban (approved):
The ticket mentions that both the crtl and var banners are to have the yellow icon but maybe that was discussed already. Otherwise all approved 🥳

moiikana:
:facepalm: thx

PR #349, opened 2020-09-14
Prepare campaign for B20_WMDE_Test_06

Abban:
https://phabricator.wikimedia.org/T262790

### DIFF
--- a/desktop/styles/styles_var.pcss
+++ b/desktop/styles/styles_var.pcss
@@ -2,12 +2,12 @@
 
 @import '../../shared/components/BannerTransition.pcss';
 
-@import './Banner.pcss';
+@import './Banner_var.pcss';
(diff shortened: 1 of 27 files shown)


desktop/styles/styles_var.pcss
```diff
@@ -2,12 +2,18 @@
...
 @import './ProgressBar.pcss';
 @import './BackLink.pcss';
+@import './Tooltip.pcss';
 @import '../../shared/components/ui/FundsModal.pcss';
+
+.wmde-banner {
+	p {
+		margin: 0 0 8px;
+	}
+}
```
moiikana: the placing of this is a little hidden here .. even if it's used at many places, maybe Banner.pcss or sth would be a better place.

PR #368, opened 2020-09-30
Prepare campaign for C20_WMDE_Mobile_Test_03

Abban:
https://phabricator.wikimedia.org/T263397
https://phabricator.wikimedia.org/T263395


mobile/components/ui/form/DonationFormWithHeaders_var.jsx
```diff
@@ -0,0 +1,225 @@
...
+
+		if ( e.target.value !== AddressType.FULL.value ) {
+			addDisabledPaymentMethod( PaymentMethods.DIRECT_DEBIT.value );
+		} else {
+			removeDisabledPaymentMethod( PaymentMethods.DIRECT_DEBIT.value );
+		}
+	};
+
+	const getFormAction = () => {
+		const piwick = '?piwik_campaign=' + props.campaignName + '&piwik_kwd=' + props.bannerName;
```
gbirke: Could you change the name of this variable to `tracking`?

mobile/components/ui/form/DonationFormWithHeaders_var.jsx
```diff
@@ -0,0 +1,225 @@
...
+			setDisabledAddressTypes( [ AddressType.EMAIL.value, AddressType.NO.value ] );
+		} else {
+			setDisabledAddressTypes( [] );
+		}
+	};
+
+	const onChangeAddressType = e => {
+		setAddressType( e.target.value );
+
+		if ( e.target.value !== AddressType.FULL.value ) {
```
gbirke: could you switch this condition around to make it easier to read?

mobile/components/ui/form/DonationFormWithHeaders_var.jsx
```diff
@@ -0,0 +1,225 @@
...
+		} else if ( paymentMethod === PaymentMethods.BANK_TRANSFER.value ) {
+			return Translations[ 'submit-label-bank-transfer' ];
+		} else {
+			return Translations[ 'submit-label-default' ];
+		}
+	};
+
+	return <div className="form">
+		<form method="post" name="donationForm" className="form__element"
+			action={ getFormAction() }>
```
gbirke, line 143: While this works, calling functions feels inconsistent to me. Maybe make formAction a reactive state and change it in `onChangeAddressType`?

mobile/components/ui/form/DonationFormWithHeaders_var.jsx
```diff
@@ -0,0 +1,225 @@
...
+				</button>
+			</div>
+
+			<Footer/>
+
+			<SubmitValues
+				amount={ props.formatters.amountForServerFormatter( numericAmount ) }
+				interval={ paymentInterval }
+				paymentType={ paymentMethod }
+				impressionCounts={ props.impressionCounts }
```
gbirke, line 232: you need to add the `addressType={ addressType }` property here.

`SubmitValues` already has that property, no need to change it.

mobile/components/ui/form/DonationFormWithHeaders_var.jsx
```diff
@@ -0,0 +1,225 @@
...
+	const [ paymentInterval, setInterval, intervalValidity, setIntervalValidity ] = useInterval( null );
+	const [ paymentMethod, setPaymentMethod, paymentMethodValidity, setPaymentMethodValidity ] = usePaymentMethod( null );
+	const [
+		{ numericAmount, amountValidity, selectedAmount, customAmount },
+		{ selectAmount, updateCustomAmount, validateCustomAmount, setAmountValidity }
+	] = useAmountWithCustom( null, props.formatters.customAmountInputFormatter );
+	const [ addressType, setAddressType, addressTypeValidity, setAddressTypeValidity ] = useAddressType( null );
+	const [ disabledIntervals, setDisabledIntervals ] = useState( [] );
+	const [ disabledPaymentMethods, setDisabledPaymentMethods ] = useState( [] );
+	let [ disabledAddressTypes, setDisabledAddressTypes ] = useState( [] );
```
gbirke:
```suggestion
	const [ disabledAddressTypes, setDisabledAddressTypes ] = useState( [] );
```

gbirke (changes requested):

mobile/components/ui/form/DonationFormWithHeaders_var.jsx
Abban: I think the issue with that is the initial value when the ref is created. It would have to call the same function to get it eg:

`const [ formAction, setFormAction ] = useState( getFormAction() );`

Then if we did that we'd also do it in `onChangeAddressType`:

```js
const onChangeAddressType = e => {
	setAddressType( e.target.value );
	setFormAction( getFormAction() ); 
...
);
```
So what we end up with is the same function and an extra ref.

gbirke: After revisiting this, I think it's fine. I think my brain was in Vue mode when I wrote my comment, where reactive state would have been better. In this case, it's totally fine.

gbirke (approved):

PR #381, opened 2020-10-13
C20_WMDE_Test_08

moiikana:
based on desktop banner 07 ctrl (ctr leads to provadd=0, var leads to provadd=1 in the fundraising app)
For 8: ctrl and var lead to provadd=0
https://phabricator.wikimedia.org/T265252

### DIFF
--- a/desktop/components/ui/form/DonationFormWidthInfoHeader.jsx
+++ b/desktop/components/ui/form/DonationFormWidthInfoHeader.jsx
@@ -13,7 +13,6 @@ import useInterval from '../../../../shared/components/ui/form/hooks/use_interva
 import usePaymentMethod from '../../../../shared/components/ui/form/hooks/use_payment_method';
 import { amountMessage, validateRequired } from '../../../../shared/components/ui/form/utils';
 import SubmitValues from './SubmitValues';
-import { BannerType } from '../../../Banner';
 
 export default function DonationForm( props ) {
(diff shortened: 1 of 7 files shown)


desktop/components/ui/form/DonationFormWidthInfoHeader.jsx
```diff
@@ -42,7 +42,8 @@ export default function DonationForm( props ) {
 	const formActionParams = {
 		piwik_campaign: props.campaignName,
 		piwik_kwd: props.bannerName,
-		provadd: props.bannerType === BannerType.CTRL ? 0 : 1
+		// TODO do we need this check more often during campaign2020? then leave it in
+		provadd: props.bannerType === BannerType.CTRL ? 0 : 0
```
gbirke: You can remove this, the bucket test campaign is configured with `param_only`, which means if the param is missing, it will choose the "old style" automatically. If FunOps decides to make the new style default, we can always add back `provadd: 1`

gbirke (approved):

desktop/styles/Infobox.pcss
```diff
@@ -1,5 +1,6 @@
 .wmde-banner {
 	.infobox {
+		height: 100%;
```
Abban, line 3: I suspect this might not work in IE11 as it has issues with `height: 100%`. It might need to be made flex. 🦾

Abban (approved):

PR #388, opened 2020-10-20
[DNM] Improve custom amount

gbirke:


gbirke:
@moiikana @Abban This is just a work in progress PR to suggest how things could be changed. Please try it out and comment. I'm not happy with having to change the form component *and* the store. I'll chat with UX to see if the current behavior (falling back to old value before changing to the new one) is acceptable or if we need to implement a different one.

gbirke:
Will create a new and improved PR on top of master

PR #390, opened 2020-10-23
Document campaign parameters

gbirke:
Document how the camapign parameters work
Add Travis check to protect against checking in development campaign parameters.


Abban (approved):
All good, just added a small follow up with a diagram for how the projection works https://github.com/wmde/fundraising-banners/pull/396

moiikana (approved):

PR #400, opened 2020-10-28
C20 wmde EN desktop 01

moiikana:
- EN desktop 01 is a text test: introduces new "late campaign sentence" that only shows up after 2 campaign days

https://phabricator.wikimedia.org/T266102


Abban:
Think it would be easy-ish to pull the cleaned up styles over from the current desktop version? Then the easy select footer could also be included.

Edit: I copied and pasted the styles in and it's mostly working, It seems to only need the custom amount euro symbol swapped to the other side

moiikana:
rebased to master,
still needs new **use of funds**

Abban (approved):

moiikana:
integrated campaigndaysentence in VAR according to this comment:
https://phabricator.wikimedia.org/T266102#6599929

PR #403, opened 2020-10-30
Add donation target to progress bar

Abban:
https://phabricator.wikimedia.org/T266548

Suggestions for the option name are more than welcome!


gbirke (approved):
I'm fine with the name, describes exactly what it does

PR #404, opened 2020-10-30
Fix slider for C20_WPDE_Mobile_01

moiikana:
- slider animation can start right away because there is no presenting animation (slide in) for the minibanner on WPDE

https://phabricator.wikimedia.org/T265253

### DIFF
--- a/shared/components/BannerTransition.jsx
+++ b/shared/components/BannerTransition.jsx
@@ -60,6 +60,9 @@ export default class BannerTransition extends Component {
 	};
 
 	onTransitionEnd = () => {
+		if ( this.state.transitionPhase === FINISHED ) {
+			return;
+		}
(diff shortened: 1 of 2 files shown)


gbirke:
While the fix works, I don't understand *why* it works and would like to look a little bit more deeply, so please don't merge yet ...

gbirke (approved):
Your fix did not work for me, but I found the reason why the slider stopped - onTransitionEnd fired twice, ~2 seconds apart (no idea why). I've added a guard clause against that.

moiikana:
> Your fix did not work for me, but I found the reason why the slider stopped - onTransitionEnd fired twice, ~2 seconds apart (no idea why). I've added a guard clause against that.

thanks for checking properly, I now noticed too ..why tf did it work for me earlier :facepalm:

PR #405, opened 2020-10-30
C20_WMDE_Mobile_Test_04

Abban:
https://phabricator.wikimedia.org/T266754
https://phabricator.wikimedia.org/T266753


moiikana (approved):
I'm a bit scared about how splitting the css into folders + subfolders (for the themes) will turn out, I hope it won't get messy. But I'd also give it a try, because last year I also felt that it really was a pitty that we couldn't extract certain "themes" (that later came back in 1-2 tests) and hat to pick them out of the commit history somehow, I think.

PR #411, opened 2020-11-02
Adjust custom amount for placeholder visibility on English desktop banners

Abban:
https://phabricator.wikimedia.org/T266887

This adjusts the minimium width of the custom amount field on smaller screens to make sure the placeholder stays visible.

### DIFF
--- a/english/styles/SelectGroup.pcss
+++ b/english/styles/SelectGroup.pcss
@@ -154,6 +154,7 @@ $select-group-line-height: 30px;
 			border: 1px solid $color-gray-light;
 			box-sizing: border-box;
 			width: 100%;
+			min-width: 110px;
 			margin: 2px 0;
 			line-height: 25px;
 			height: 26px;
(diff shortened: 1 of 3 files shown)


shared/components/ui/form/SelectGroup.jsx
```diff
@@ -30,7 +30,10 @@ export function SelectGroup( props ) {
 
 		<div className="select-group">
 			{ props.selectionItems.map( ( { value, label } ) => (
-				<label className={ 'select-group__option' } key={ value }>
+				<label className={ classNames(
+					'select-group__option',
+					`select-group__option-${ props.fieldname }-${value.replace( ' ', '-' )}`,
```
gbirke:
```suggestion
					`select-group__option--${ props.fieldname }-${value.replace( ' ', '-' )}`,
```

english/styles/SelectGroup.pcss
```diff
@@ -243,4 +244,11 @@ $select-group-line-height: 30px;
 			}
 		}
 	}
+
+	.select-group__option-select-interval-6 {
```
gbirke:
```suggestion
	.select-group__option--select-interval-6 {
```

gbirke (approved):
BEM modifiers should be separated with two dashes

PR #412, opened 2020-11-02
C20_WMDE_mobile_en_test_01

moiikana:
- first mobile EN test 2020
- introduces rightText changes for progress bar
- progress bar animation starts

- VAR: introduces address provision options (where already
implemented in DE mobile 03 2020)

https://phabricator.wikimedia.org/T266756


campaign_info.toml
```diff
@@ -39,19 +39,19 @@ tracking = "org-mob03-201013-var"
...
-campaign_tracking = "mob_en02-ba-191206"
+campaign_tracking = "mob_en01-ba-201103"
 preview_link = "/wiki/Wikipedia:Hauptseite?devbanner={{banner}}&useskin=minerva&banner=B17WMDE_webpack_prototype"
 wrapper_template = "wikipedia_org"
 
 [mobile_english.banners.ctrl]
 filename = "./mobile_english/banner_ctrl.js"
-pagename = "B19WMDE_mob_en02_191206_ctrl"
-tracking = "org-mob_en02-191206-ctrl"
+pagename = "B19WMDE_mob_en01_201103_ctrl"
```
Abban: This should probably be B20?

shared/components/ui/Infobox.jsx
```diff
@@ -19,6 +20,7 @@ export default function Infobox( { campaignParameters, campaignProjection, forma
 		endOfDay( campaignParameters.endDate )
 	);
 	const campaignDaySentence = new CampaignDaySentence( campaignDays, Translations );
+	const visitorsVsDonorsSentence = new VisitorsVsDonorsSentence( campaignParameters, campaignProjection, campaignDays, Translations );
```
Abban: This is much better than the implementation I did in the [.de banner](https://github.com/wmde/fundraising-banners/pull/402/files#diff-ede34ae6ca87afb22d9a78f593fc2397722bdf86f33bc880b511fe1ee54a1249R7).

As a rule I would pass in only the things the VisitorsVsDonorsSentence class needs to know about though, it doesn't need to know anything about the other objects:

```js
const visitorsVsDonorsSentence = new VisitorsVsDonorsSentence(
    campaignParameters.millionImpressionsPerDay,
    campaignProjection.getProjectedNumberOfDonors(),
    campaignDays.getDaysSinceCampaignStart(),
    Translations[ 'visitors-vs-donors-sentence' ]
);
```

Abban (changes requested):
Looking good!

One other thing I noticed apart from the comments is that the progressbar slide should be the first one in each banner.

campaign_info.toml
moiikana: goddammit

shared/components/ui/Infobox.jsx
moiikana: aah.. that's much easier indeed. I'll apply your suggestions

Abban (approved):
Looks good! Probably should wait and see how we're adding use of funds before merging though.

PR #413, opened 2020-11-02
Adjust custom amount for placeholder visibility on DE desktop banners

Abban:
https://phabricator.wikimedia.org/T266887

This adjusts the minimium width of the custom amount field on smaller screens to make sure the placeholder stays visible.


gbirke (changes requested):
Same comments as in #411, please also add a prefix so we can distinguish the PRs, I was a bit confused by the identical PR names

PR #397, opened 2020-10-26
[DNM] Add new use of funds to banners

gbirke:
See https://phabricator.wikimedia.org/T266114

Changes will be merged with individual PRs into banners

PR #414, opened 2020-11-02
Add new use of funds to desktop 09 banner

gbirke:
As specified in https://phabricator.wikimedia.org/T257185
With applied requested content changes from https://phabricator.wikimedia.org/T266114
Improved display on small desktop viewport width:
- Min width of container is 800px
- Don't do two columns when viewport width < 1024
- Adapt bar font size

This is a squashed & adapted version of #397


Abban (approved):
The only thing I noticed is that the close button on the modal scrolls with the content, I think that should be fixed position?

moiikana (approved):

PR #416, opened 2020-11-02
Add use of funds to wikipedia.de desktop 01 banner

gbirke:


gbirke:
Base was rebased, will redo this

PR #421, opened 2020-11-03
Fix mobile EN 01 slides

moiikana:
changes are done according to this comment:
>
- The slide containing the progress bar should be the first.
- The visitors-vs-donors sentence is part of one slide. That's fine, but we should use the placeholder, so it is not displayed on the first campaign day.

https://phabricator.wikimedia.org/T266756#6599732

### DIFF
--- a/mobile_english/components/Slides.jsx
+++ b/mobile_english/components/Slides.jsx
@@ -1,20 +1,21 @@
 // eslint-disable-next-line no-unused-vars
 import { h } from 'preact';
 
-export default function Slides( { formattedGoalDonationSumNumeric, currentDayName, numberOfDonors, campaignParameters, progressBar } ) {
+export default function Slides( { formattedGoalDonationSumNumeric, currentDayName, visitorsVsDonorsSentence, progressBar } ) {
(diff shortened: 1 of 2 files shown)


mobile_english/components/Slides.jsx
```diff
@@ -1,19 +1,18 @@
...
 		<div className="carousel-cell">
-			<p>Our fundraising appeal is displayed over { campaignParameters.millionImpressionsPerDay } million times a day,
-				but currently only { numberOfDonors } people have donated. The price of a coffee is all we need.</p>
+			<p>To all our readers in Germany. We will get straight to the point:
+				This { currentDayName } we ask you to protect Wikipedia's independence.</p>
 		</div>
 		<div className="carousel-cell">
-			<p className="goal-headline">Our donation target: { formattedGoalDonationSumNumeric } million Euros</p>
-			{ progressBar }
+			<p>{ visitorsVsDonorsSentence }</p>
```
Abban: This will cause a blank slide to appear on day 1.

Abban (approved):
All good, just needs an if statement around the visitorsVsDonersSentence

PR #428, opened 2020-11-04
Add right margin to form

Abban:
Also fixes js error thrown on submit

### DIFF
--- a/wikipedia.de/desktop/Banner.jsx
+++ b/wikipedia.de/desktop/Banner.jsx
@@ -126,7 +126,7 @@ export default class Banner extends Component {
 							impressionCounts={ props.impressionCounts }
 							onFormInteraction={ this.onFormInteraction }
 							customAmountPlaceholder={ props.translations[ 'custom-amount-placeholder' ] }
-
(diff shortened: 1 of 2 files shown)


wikipedia.de/desktop/Banner.jsx
```diff
@@ -126,7 +126,7 @@ export default class Banner extends Component {
 							impressionCounts={ props.impressionCounts }
 							onFormInteraction={ this.onFormInteraction }
 							customAmountPlaceholder={ props.translations[ 'custom-amount-placeholder' ] }
-
+							onSubmit={ () => {} }
```
moiikana, line 129: shouldn't this be:
```suggestion
							onSubmit={props.onSubmit}
```

the onSubmit comes from the banner_presenter and seems to do the tracking, not sure, but should also be applied to WPDE?

Abban: The DonationForm component posts the actual html form to spenden so I think it's okay. All the empty function does is block the error that's thrown right before the post happens.

moiikana (approved):

PR #441, opened 2020-11-06
Center slider progressbar vertically

moiikana:
- centers the progressbar on MobileWPDE + DE mobile vertically
- does not need centering on EN mobile because the sliders
are very small there

https://phabricator.wikimedia.org/T267385


Abban (approved):
This is an interesting solution as vertically centering it with flex isn't possible due to the slideshow styles. I think this is the best we can get it.

PR #452, opened 2020-11-10
C20 wmde test 11

Abban:
Config: https://phabricator.wikimedia.org/T267527
Setup: https://phabricator.wikimedia.org/T267518

Text changes in both banners, and a new event when the use of funds call to action is clicked

### DIFF
--- a/desktop/Banner.jsx
+++ b/desktop/Banner.jsx
@@ -92,6 +92,11 @@ export class Banner extends Component {
 		this.setState( { isFundsModalVisible: !this.state.isFundsModalVisible } );
 	};
 
+	fundsModalDonate = () => {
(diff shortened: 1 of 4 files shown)


desktop/Banner.jsx
```diff
@@ -92,6 +92,11 @@ export class Banner extends Component {
 		this.setState( { isFundsModalVisible: !this.state.isFundsModalVisible } );
 	};
 
+	fundsModalDonate = () => {
+		this.props.trackingData.tracker.trackBannerEvent( 'funds-modal-donate-clicked', 0, 0, this.props.trackingData.bannerClickTrackRatio );
```
moiikana, line 96: in all other banners it's currently called 'application-of-funds-shown', maybe that should be unified

Abban: This event is specifically for the funds modal call to action button

moiikana: oops, then I didn't say anything :+1:

moiikana (approved):

PR #456, opened 2020-11-11
C20 wmde en test 02

Abban:
https://phabricator.wikimedia.org/T267601

Add 2 step form to var banner


moiikana:
2 things I noticed: 
- at around 1500px viewport width the CTRL banner seems to have a smaller form than desktop-en-01 ctrl.
- at 1214px the CTRL bannertext has no linebreak after the headline, the old en01 ctrl has

shared/messages/de.js
```diff
@@ -73,7 +73,7 @@ const Translations = {
 	'submit-label-bank-transfer': 'Weiter zur Bankverbindung',
 	'submit-label-short': 'Weiter',
 	'submit-label-default': 'Weiter zur Adresseingabe',
-	'address-type-notice-nein': 'Ihre Kontaktdaten benötigen wir für die Bestätigungsemail und Ihre Spendenquittung. ' +
+	'address-type-notice-negative': 'Ihre Kontaktdaten benötigen wir für die Bestätigungsemail und Ihre Spendenquittung. ' +
```
moiikana, line 76: that line of texts already exists with the key 'address-type-info-message' in the same file, that should be used instead

shared/messages/en.js
```diff
@@ -78,7 +78,8 @@ const Translations = {
 	'submit-label-credit-card': 'Proceed with credit card',
 	'submit-label-sofort': 'Proceed with Sofort-Überweisung',
 	'submit-label-bank-transfer': 'Proceed with bank transfer',
-	'submit-label-default': 'Proceed with the donation'
+	'submit-label-default': 'Proceed with the donation',
+	'address-type-notice-disabled': 'A postal address is necessary for donating by direct-debit.'
```
moiikana, line 82: same text already exists here with key 'address-type-notice-direct-debit'

shared/messages/de.js
Abban: We also have an `address-type-notice-disabled` item in that language sheet. This one isn't actually being used in this test, I only renamed it as I noticed the key wasn't in English.

shared/messages/en.js
Abban: Yeah, one is for the field and one is for the notice on form step 2

moiikana (approved):

PR #457, opened 2020-11-11
Make use of funds translatable and reusable

gbirke:
This is for https://phabricator.wikimedia.org/T267628 and https://phabricator.wikimedia.org/T266126


gbirke:
The graphic might change, necessitating an npm update of the content repo package, but apart from that, I consider this to be done.

shared/components/ui/use_of_funds/FundsContent.jsx
```diff
@@ -1,134 +1,96 @@
...
-						Also für Sie – und alle Menschen weltweit.
-					</div>
-				</div>
-				{props.children}
+export default function FundsContent( props ) {
+	const onCallToAction = props.onCallToAction ? props.onCallToAction : props.toggleFundsModal;
+	const content = props.useOfFundsText;
+	const organizationClassLookup = new Map( Object.entries( content.orgchart.organizationClasses ) );
+	const getHighlightClassName = part => organizationClassLookup.has( part ) ? `use_of_funds__org use_of_funds__org--${organizationClassLookup.get( part )}` : '';
+	const highlightOrganizations = text => splitStringAt( Array.from( organizationClassLookup.keys() ), text ).map( part =>
```
Abban, line 13: Because there's whitespace being added in the span, this causes a little issue in the English version where the highlighted `Wikimedia Foundation` text is at the end of a sentence and there's a space before the `.`. It might be simpler to add the spans directly in the translations (`<span class="highlight-orange"></span>`), or put in template tags `{{wmfg}}` that we can run through `string.replace`

gbirke: I'm not sure that will work. If you look at the output of `splitStringAt`, you'll see that the whitespace inside the strings is correct, but react removes it due to https://github.com/facebook/react/issues/1643 Changing the replacement mechanism won't save us from React stripping whitespace in span elements (but only in some cases).

I have found a fugly workaround (looking for punctuation in `part`) for our special case.

moiikana (approved):
only needs a rebase + linting fix and then all good from my side

PR #460, opened 2020-11-16
Fix banner displaying after closing WPDE

moiikana:
After closing the banner,
the banner showed up after the presenting animation time.

https://phabricator.wikimedia.org/T267899

### DIFF
--- a/wikipedia.de/desktop/Banner.jsx
+++ b/wikipedia.de/desktop/Banner.jsx
@@ -44,14 +44,13 @@ export default class Banner extends Component {
 
 	componentDidMount() {
 		this.props.registerDisplayBanner(
-			() => {
-				this.setState( { displayState: VISIBLE } );
-			}
+			() => {}
 		);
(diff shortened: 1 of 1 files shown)


wikipedia.de/desktop/Banner.jsx
```diff
@@ -44,9 +44,7 @@ export default class Banner extends Component {
 
 	componentDidMount() {
 		this.props.registerDisplayBanner(
-			() => {
-				this.setState( { displayState: VISIBLE } );
-			}
+			() => {}
 		);
 		this.props.registerResizeBanner( this.adjustSurroundingSpace.bind( this ) );
```
Abban, line 49: All good, but can you re-add `this.setState( { displayState: VISIBLE } );` at the bottom of the `componentDidMount` method? Even though we're not making this banner hidden in the styles it'll make it more consistent with how the other banners work.

Abban (approved):

PR #466, opened 2020-11-17
Make localstorage access more resilient

gbirke:
Avoid throwing errors when accessing localstorage

Bug: https://phabricator.wikimedia.org/T267267

### DIFF
--- a/shared/local_impression_count.js
+++ b/shared/local_impression_count.js
@@ -22,10 +22,7 @@ export class LocalImpressionCount {
 		try {
 			return window.localStorage.getItem( name ) || defaultValue;
 		} catch ( e ) {
-			if ( e.name === 'NS_ERROR_FILE_CORRUPTED' ) {
-				return defaultValue;
-			}
-			throw e;
+			return defaultValue;
 		}
 	}
 
@@ -41,10 +38,7 @@ export class LocalImpressionCount {
 			window.localStorage.setItem( 'fundraising.overallCount', this.overallCount.toFixed( 0 ) );
 			window.localStorage.setItem( 'fundraising.bannerCount', this.bannerName + '|' + this.bannerCount );
 		} catch ( e ) {
-			if ( e.name === 'NS_ERROR_FILE_CORRUPTED' ) {
-				return;
-			}
-			throw e;
+			// Don't throw localStorage exceptions
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


shared/local_impression_count.js
```diff
@@ -18,14 +18,32 @@ export class LocalImpressionCount {
 		}
 	}
 
+	getItem( name, defaultValue ) {
+		try {
+			return window.localStorage.getItem( name ) || defaultValue;
+		} catch ( e ) {
+			if ( e.name === 'NS_ERROR_FILE_CORRUPTED' ) {
```
Krinkle: This will also throw in private browsing mode, or when the interface is disallowed for other reasons by browsers. I don't think a general distinction is feasible in browsers between an ignorable and significant erorr here. There are no significant errors possible in localStorage. The only reason for getItem to throw is if something is wrong unrelated to the specific key or value and thus always ignorable. mw.storage and other JS libs follow that direction as well.

shared/local_impression_count.js
```diff
@@ -18,14 +18,32 @@ export class LocalImpressionCount {
...
 		if ( !window.localStorage ) {
 			return;
 		}
-		window.localStorage.setItem( 'fundraising.overallCount', this.overallCount.toFixed( 0 ) );
-		window.localStorage.setItem( 'fundraising.bannerCount', this.bannerName + '|' + this.bannerCount );
+		try {
+			window.localStorage.setItem( 'fundraising.overallCount', this.overallCount.toFixed( 0 ) );
+			window.localStorage.setItem( 'fundraising.bannerCount', this.bannerName + '|' + this.bannerCount );
+		} catch ( e ) {
+			if ( e.name === 'NS_ERROR_FILE_CORRUPTED' ) {
```
Krinkle: (idem)

shared/local_impression_count.js
```diff
@@ -6,9 +6,9 @@ export class LocalImpressionCount {
 		if ( !window.localStorage ) {
```
Krinkle: This can also throw an exception on itself. In private browsing, the access to the object is disallowed. This could, for example, be handled as part of the try-catch instead.

In general there is no need to account for the interface being unsuported (undefined) since there are no browsers in which MW ships JS code that don't have this interface. It is explicitly one of the three feature-tests MW checks for in its startup module based on a safe property lookup (not value access). – [startup.js](https://github.com/wikimedia/mediawiki/blob/4d6e4aa0f20cf2ae3a3f6ced6bcbfea50c82e2c8/resources/src/startup/startup.js)

PR #483, opened 2020-11-30
C20_WMDE_Test_19 [DNM, old faulty flickery implementation]

moiikana:
-based on test 18 ctrl
- VAR: adds sliding infobox to desktop
https://phabricator.wikimedia.org/T268969


desktop/components/BannerText.jsx
```diff
@@ -26,23 +28,39 @@ export default function BannerText( props ) {
...
-					Sie zu dem kleinen Kreis gehören, die bereits gespendet haben, danken wir Ihnen sehr herzlich.
-					Wikipedia wird durch Spenden von durchschnittlich 22,81&nbsp;€ finanziert. Doch schon mit einem
-					kleinen Beitrag kann Wikipedia sich auch in Zukunft erfolgreich entwickeln. { visitorsVsDonorsSentence } Die meisten Menschen spenden,
-					weil sie Wikipedia nützlich finden. Hat Wikipedia Ihnen in diesem Jahr Wissen im Wert von 5&nbsp;€ geschenkt?
-					Dann nehmen Sie sich doch bitte eine Minute Zeit und geben Sie etwas zurück. <span className="optional-text text-l">Zeigen
-			Sie den Freiwilligen, die Ihnen verlässliche und neutrale Informationen zur Verfügung stellen,
-			dass Sie ihre Arbeit wertschätzen.</span> Vielen Dank!
-				</p>
+				<a href="#" className="banner-slider-previous" onClick={ bannerSliderPrevious }><ChevronLeftIcon/></a>
+				<div className="mini-banner-carousel">
```
gbirke, line 32: This name is not strictly correct, since we don't distinguish between mini and full banner on desktop. But let's keep the name for now, so we can compare the CSS files better.

gbirke (approved):
Thanks for cleaning up the unused files!

desktop/components/BannerText.jsx
Abban: The reason it was kept is it's hardcoded into the Slider class initializer, it was easier to leave it than update every banner. If sliders become the norm then we should refactor.

Abban:
The close button on the var banner needs a z-index bump as it currently can't be clicked on lower sizes.

PR #487, opened 2020-12-02
Do not show banner if media-viewer got opened

moiikana:
If an image is clicked and the mediaviewer is open, the main content of the page
gets set invisible. This change checks if the main content is visible
and the banner can be rendered there properly.

https://phabricator.wikimedia.org/T268918

### DIFF
--- a/shared/banner_presenter.js
+++ b/shared/banner_presenter.js
@@ -1,6 +1,10 @@
 import SizeIssueIndicator from './track_size_issues';
 import { getSkinAdjuster } from './skin';
-import { mediaWikiIsShowingContentPage, onMediaWiki } from './mediawiki_checks';
+import {
+	mediaWikiIsShowingContentPage,
+	mediaWikiMainContentIsHiddenByLightbox,
+	onMediaWiki
+} from './mediawiki_checks';
 import { createElement, render } from 'preact';
 import InterruptibleTimeout from './interruptible_timeout';
 import { VIEWPORT_TRACKING_IDENTIFIER, VIEWPORT_TRACKING_SUBMITTED_EVENT_IDENTIFIER } from './event_logging_tracker';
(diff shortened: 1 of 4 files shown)


gbirke (changes requested):
Unfortunately this won't work as expected. The other "Observers" in the SkinAdjuster observe events (user focusing on search field, user clicking on edit button).  If we decide to stay with the event-based approach, `addMediaOverlayObserver` should actively look for the "mediaViever is activated" event. Unfortunately, the MediaViwever is a black box that doesn't have events/hooks that our code could attach on. Instead, it uses a [complex detection logic](https://github.com/wikimedia/mediawiki-extensions-MultimediaViewer/blob/72636057d8997da3a0523624341afc30a703f1f1/resources/mmv.bootstrap/mmv.bootstrap.js#L181-L249) to attach itself to most image links on the page. 
IMHO, the next best step here would be to observe side effects, e.g, the `body` element gaining the class `mw-mmv-lightbox-open`. You can do this witha MutationObserver class, as explained here https://stackoverflow.com/a/19401707/130121

Another way of solving this would be a new skinAdjuster method, e.g. named `platformStateAllowsBannerDisplay` or `bannerIsStillAllowedToDisplay` that gets called by the `BannerPresenter` before it finally displays the banner. There you could put the check if the element is visbile (or if `body` has the `mw-mmv-lightbox-open` class) and maybe even migrate the searchObserver code there, to allow the banner to display when the user has focused and unfocused the search field.

shared/banner_presenter.js
```diff
@@ -32,7 +36,7 @@ export default class BannerPresenter {
 		}
 		const sizeIssueIndicator = new SizeIssueIndicator( sizeIssueThreshold );
 
-		if ( onMediaWiki() && !mediaWikiIsShowingContentPage() ) {
+		if ( !mediaWikiIsShowingContentPage() || mediaWikiMainContentIsHiddenByLightbox() ) {
```
moiikana, line 39: `mediaWikiMainContentIsHiddenByLightbox()` is checked at 2 places in the `banner_presenter.js`, because it is needed for 2 cases:
1) mediaviewer is open, then page refresh (f5), banner tries to load in the background
2) mediaviewer gets opened during banner display delay, banner tries to load in the background

gbirke (approved):

PR #493, opened 2020-12-08
Adjust styles for late_progress stage

moiikana:
https://phabricator.wikimedia.org/T269612

### DIFF
--- a/desktop/styles/ProgressBar.pcss
+++ b/desktop/styles/ProgressBar.pcss
@@ -61,7 +61,7 @@ $progress-bar-height: 30px;
 			white-space: nowrap;
 
 			color: $color-black;
-			margin-left: -40px;
+			line-height: 2.5em;
 		}
 		&__pointer_tip {
 			width: 5px;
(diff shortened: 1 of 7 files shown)


gbirke:
Thank you so much for tackling this! I don't know if this is already meant to be reviewed, but I had a quick look and noticed the following things:

* desktop VAR is not changed yet, but that's probably because of the new design
* In the desktop wp.de banner, the text is a bit too close to the banner border  
Bildschirmfoto 2020-12-08 um 18 15 39

* In the mobile (ipad both languages, mobile EN) the border radius of the bar does not fit with the border radius of its frame. I think this is not intentional, it's just that at some point we changed the outer one radius to make it look nicer and forgot how the two radii will look when they are adjacent. But I think @KaiNissen has to chime in here to tell us what the final design should be.  
Bildschirmfoto 2020-12-08 um 18 16 37

moiikana:
thanks @gbirke, I hadn't been working on the WPDE banners at the time of your comment, that is fixed now.
for the filling shape: according to Kai this is ok I think

Abban (approved):

PR #518, opened 2021-01-27
Use Of Funds Company Budget Currency Symbol Fix

Abban:
https://phabricator.wikimedia.org/T272834

We were being a little too smart with the budget string localisation. I made it more dumb and moved it into the content repo.


package.json
```diff
@@ -18,25 +18,25 @@
   "repository": "https://github.com/wmde/fundraising-banners",
   "dependencies": {
     "classnames": "^2.2.6",
-    "flickity": "^2.2.2",
+    "flickity": "^2.2.1",
     "format-number": "^3.0.0",
-    "fundraising-frontend-content": "github:wmde/fundraising-frontend-content#production",
-    "keen-slider": "^5.4.0",
+    "fundraising-frontend-content": "github:wmde/fundraising-frontend-content#test",
```
gbirke: How can we make sure to remember to switch back to production again?

gbirke (approved):
Thank you for simplifying this! Being "not clever" pays off here.

package.json
Abban: I wonder if it's possible to specify #test as a dev dependency and #production on build?

moiikana (approved):

PR #645, opened 2021-06-29
Introduce keen-slider library for mobile banners

moiikana:
Adapts properties / configurations from flickety library to the new keen-slider library.

- [x] make design look like the former mobile banners
  - [x] implement autoplay functionality that pauses on interaction
  - [x] add dots to navigate
  - [x] check ctrl+var versions
- [x] remove/hide progressbar for now
- [x] check if slide tracking+counting works correctly

https://phabricator.wikimedia.org/T273422


gbirke (changes requested):
See my changes in #651 
Instead of commenting in this PR, I've made individual commits with comments there, so you can understand better what I mean.
Please don't do a merge commit of #651, do a "squash and merge" instead, that PR is just a fancy way of doing a review with suggestions ;-)

gbirke (approved):

PR #652, opened 2021-07-07
C21_WMDE_Test_02

moiikana:
- based on desktop test 21 var
https://de.wikipedia.org/wiki/Wikipedia:Hauptseite?banner=B20_WMDE_Test_21_var

TODOs:
- [x]  check tracking calls
- [x] fix spacing on L breakpoint (right margin of the slider)

https://phabricator.wikimedia.org/T286180


desktop/styles/Slider_var.pcss
```diff
@@ -0,0 +1,109 @@
...
+								@media ( min-width: $breakpoint_m ) {
+									font-size: 16px;
+								}
+								@media ( min-width: $breakpoint_l ) {
+									font-size: 17px;
+								}
+							}
+							.headline {
+								font-size: 22px;
+								line-height: 26px;
```
Abban: The headers are a little squashed when they break to a new line because the line height is the same as the font size, if you change the line-height to `1.3` it'll always have decent spacing no matter the font-size.

desktop/components/Banner_var.jsx
```diff
@@ -28,36 +26,23 @@ const BannerVisibilityState = Object.freeze( {
...
-} );
-
-export class Banner extends Component {
+export default class Banner extends Component {
 
 	static propTypes = {
 		/** callback when banner closes */
 		onClose: PropTypes.func,
-		/** callback when banner gets submitted */
-		onSubmit: PropTypes.func,
```
Abban: It looks like `this.props.onSubmit();` is being called on line 113, so you probably still need this propType declaration. The submit is firing 2 submit events one with the window dimensions and one with the slider info, I don't remember if this is the desired functionality.

desktop/components/Banner_var.jsx
```diff
@@ -28,36 +26,23 @@ const BannerVisibilityState = Object.freeze( {
...
-		/** */
+		/** Callback to register a displayBanner function with the BannerPresenter */
 		registerDisplayBanner: PropTypes.func.isRequired
 	}
-
 	bannerRef = createRef();
 	slideshowRef = createRef();
 	contentRef = createRef();
+	ref = createRef();
```
Abban: I don't think this is being used anywhere? Also I can see `bannerRef`, `slideshowRef` and `contentRef` being added to elements, but they don't look like they're being used for anything.

desktop/components/Banner_var.jsx
```diff
@@ -175,38 +148,24 @@ export class Banner extends Component {
...
-		e.preventDefault();
-		this.trackBannerEvent( 'desktop-banner-expanded' );
-		this.setState( { bannerContentState: BannerContentState.FORM }, this.setContentSize );
 		this.stopSliderAutoplay();
-	}
-
-	showSlides = e => {
-		e.preventDefault();
-		this.setState( { bannerContentState: BannerContentState.SLIDES }, this.setContentSize );
+		this.setState( { showLanguageWarning: true, formInteractionSwitcher: !this.state.formInteractionSwitcher } );
```
Abban: `showLanguageWarning` is for the English banners

desktop/components/Banner.jsx
```diff
@@ -28,36 +26,23 @@ const BannerVisibilityState = Object.freeze( {
...
-		onSubmit: PropTypes.func,
-		/** */
+		/** Callback to register a displayBanner function with the BannerPresenter */
 		registerDisplayBanner: PropTypes.func.isRequired
 	}
-
 	bannerRef = createRef();
 	slideshowRef = createRef();
 	contentRef = createRef();
+	ref = createRef();
```
Abban: The unused variables are also here.

desktop/components/Banner.jsx
```diff
@@ -173,38 +148,24 @@ export class Banner extends Component {
...
-		e.preventDefault();
-		this.trackBannerEvent( 'desktop-banner-expanded' );
-		this.setState( { bannerContentState: BannerContentState.FORM }, this.setContentSize );
 		this.stopSliderAutoplay();
-	}
-
-	showSlides = e => {
-		e.preventDefault();
-		this.setState( { bannerContentState: BannerContentState.SLIDES }, this.setContentSize );
+		this.setState( { showLanguageWarning: true, formInteractionSwitcher: !this.state.formInteractionSwitcher } );
```
Abban: `showLanguageWarning` is here, too

desktop/components/Banner_var.jsx
```diff
@@ -28,36 +26,23 @@ const BannerVisibilityState = Object.freeze( {
 	CLOSED: Symbol( 'closed' )
```
Abban, line 25: Thers a MINIMISED state here that was from when we did the test that allowed people to hide the banner but not remove it. It's not used in this banner at all.

Abban (changes requested):
I think you made this a bit harder on yourself by modifying a different banner (the English one I think?) for use as this one. The tags come in really handy for this, as you could have downloaded the zip from the [C20_WMDE_Test_21 tag](https://github.com/wmde/fundraising-banners/releases/tag/C20_WMDE_Test_21-2), and replaced the desktop folder in the repo with the one from the zip. Git will then pick up the line changes.

That being said, there are some regressions in the `C20_WMDE_Test_21` banners that are much improved by the minimising banner that we should be using (eg the better defined high level banner states). If our base banner for this year will be the orange one we should look into integrating the best parts of both before starting on test 3.

Abban (approved):

PR #674, opened 2021-08-03
C21_WMDE_Test_04 (white slider vs. blue globe slider)

moiikana:
- CTRL is based on desktop 2 (white slider)
- additionally: centers slides, changes slide text, removes payment interval
- 
- VAR is based on last years blue globe slider with special closing behaviour

https://phabricator.wikimedia.org/T287353


campaign_info.toml
```diff
@@ -4,20 +4,20 @@
 
 # A Campaign with global settings
 [desktop]
-campaign_tracking = "03-ba-210802"
+campaign_tracking = "04-ba-210812"
```
KaiNissen: The start date was moved to Aug 20th. Production banners were already changed, but should be reproduced here.

KaiNissen: Due to a bug in banner height calculation the test was restarted on Aug 27. Campaign parameters were changed on production, but should be reproduced here. (campaign: `04-ba-210827`, banner: `org-04-210827-ctrl` and `org-04-210827-ctrl`)

PR #681, opened 2021-08-16
C21 wmde mobile test 01

Abban:
Ticket: https://phabricator.wikimedia.org/T287834


shared/banner_presenter.js
```diff
@@ -137,7 +137,9 @@ export default class BannerPresenter {
 			bannerDisplayTimeout.cancel();
 			if ( onMediaWiki() ) {
 				mw.centralNotice.setBannerLoadedButHidden();
+				mw.centralNotice.hideBanner();
```
gbirke: We need to check in this function if the banner is already displayed or not. If not displayed, cancel the timeout and call `setBannerLoadedButHidden`. Otherwise hide banner and adjust space. Also, this function applies both to the desktop and mobile environment. maybe we need to distinguish the behavior between them (not hiding the banner on desktop). I have asked Kai about that.

gbirke (changes requested):

moiikana (approved):

PR #683, opened 2021-08-19
Improve code style

gbirke:
Looking at the rules from https://github.com/preactjs/eslint-config-preact and applying what makes sense for our code base

Remove unused code


thank_you/components/ProgressBar.jsx
```diff
@@ -1,4 +1,3 @@
-// eslint-disable-next-line no-unused-vars
 import { h } from 'preact';
 // This component uses Component from compat because pure Preact does not support
```
Abban, line 2: IIRC all the new banners use CSS animation for the progress bar, so we might be able to remove compat

Abban (approved):

moiikana (approved):

PR #703, opened 2021-09-17
C21 wmde mobile test 02 

moiikana:
- word change on 1st slide and bannertext

https://phabricator.wikimedia.org/T291247

### DIFF
--- a/campaign_info.toml
+++ b/campaign_info.toml
@@ -6,6 +6,7 @@
 [desktop]
 campaign_tracking = "04-ba-210831"
 preview_link = "/wiki/Wikipedia:Hauptseite?devbanner={{banner}}&banner=B17WMDE_webpack_prototype"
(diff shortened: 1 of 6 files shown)


campaign_info.toml
```diff
@@ -26,25 +27,31 @@ resolution = ["800x600", "1024x768", "1280x960", "1600x1200", "1920x1200", "2560
 
 # Mobile Campaign
 [mobile]
-campaign_tracking = "mob01-ba-210901"
+campaign_tracking = "mob02-ba-210924"
```
Abban: Looks like this campaign is now going on the 27th

Abban (approved):
Keen slider styles might need to be adjusted to fix banner height issues. The fix is to add the styles that Keen adds inline onto the slides into our css so when the banner loads it already has the correct layout when doing the size calculations. The css to add is:

```css
.keen-slider__slide {
    min-width: calc( 100% + 0 );
    max-width: calc( 100% + 0 );
    transform: translate3d( 0, 0, 0 );
}
```

PR #725, opened 2021-10-12
C21 wmde test en 01

Abban:
The content for the Use of Funds will need to be merged into production before it updates in these banners. I want to deploy the app translations to test before merging them however.

Ticket: https://phabricator.wikimedia.org/T292732


moiikana:
>  It might a little awkward, but please don't scroll past this. 

in the info text sounds weird, might be some word missing, even in the banner from last year.
besides that it looks good

moiikana (approved):

PR #764, opened 2021-11-10
Fix B21_WMDE_ipad_en_01_var

gbirke:
The check for address and payment types now compares values.
Changes to address type now resets disabled payment types.
Correct BEM naming for disabled elements.
Reflect disabled elements through CSS rules

### DIFF
--- a/pad_english/components_var/ui/form/DonationForm.jsx
+++ b/pad_english/components_var/ui/form/DonationForm.jsx
@@ -37,8 +37,12 @@ export default function DonationForm( props ) {
 	};
 
 	useEffect(
-		() => setUrl( addressType !== AddressType.NO.value ? NEW_DONATION_URL : ADD_DONATION_URL ),
-		[ addressType, setUrl ]
+		() => {
+			const someAddressData = addressType !== AddressType.NO.value;
+			const paymentNeedsAddress = addressType === PaymentMethods.DIRECT_DEBIT;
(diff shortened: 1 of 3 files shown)


pad_english/components_var/ui/form/DonationForm.jsx
```diff
@@ -37,8 +37,12 @@ export default function DonationForm( props ) {
 	};
 
 	useEffect(
-		() => setUrl( addressType !== AddressType.NO.value ? NEW_DONATION_URL : ADD_DONATION_URL ),
-		[ addressType, setUrl ]
+		() => {
+			const someAddressData = addressType !== AddressType.NO.value;
+			const paymentNeedsAddress = addressType === PaymentMethods.DIRECT_DEBIT;
+			setUrl( someAddressData || paymentNeedsAddress ? NEW_DONATION_URL : ADD_DONATION_URL );
```
Abban, line 43: Don't the setDisabledX events prevent this state from happening? If the donor selects `AddressType.NO` they can't select `PaymentMethods.DIRECT_DEBIT` and vice versa. That would mean checking AddressType would be sufficient.

Abban (approved):

pad_english/components_var/ui/form/DonationForm.jsx
gbirke: Yeah, this is more of a "belt and suspenders" code where I don't rely on the correctness of the other state transitions and do my own.

gbirke:
I've rebased this branch onto current main branch to include the use-of-funds changes and re-deployed.

PR #777, opened 2021-11-26
Mobile slider interactions

Abban:
Update the proxy URLs so Wikipedia doesn't redirect them when testing on mobile devices

Add an onTouchStart event to the Slider so we can capture it the same way as onMouseDown

Ticket: https://phabricator.wikimedia.org/T296281

### DIFF
--- a/campaign_info.toml
+++ b/campaign_info.toml
@@ -28,7 +28,7 @@ resolution = ["800x600", "1024x768", "1280x960", "1600x1200", "1920x1200", "2560
 # Mobile Campaign
 [mobile]
 campaign_tracking = "mob06-ba-211120"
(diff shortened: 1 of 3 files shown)


gbirke (approved):
Thank you!

campaign_info.toml
```diff
@@ -50,7 +50,7 @@ orientation = [ "portrait", "landscape"]
 # Mobile English Campaign
 [mobile_english]
 campaign_tracking = "mob_en01-ba-211103"
-preview_link = "/wiki/Wikipedia:Main_Page?devbanner={{banner}}&useskin=minerva&banner=B17WMDE_webpack_prototype"
+preview_link = "/en-mobile/wiki/Main_Page?devbanner={{banner}}&useskin=minerva&banner=B17WMDE_webpack_prototype"
```
moiikana, line 53: omg thank you this was much needed

PR #779, opened 2021-12-02
Mobile de 07

gbirke:
https://phabricator.wikimedia.org/T296831


mobile/components/ui/form/DonationFormWithAddress.jsx
```diff
@@ -0,0 +1,244 @@
...
+						disabledOptions={ disabledPaymentMethods }
+						errorPosition={ props.errorPosition }
+					>
+						<SmsBox/>
+					</SelectGroup>
+				</div>
+			</fieldset>
+
+			<fieldset className="form__section">
+				<legend className="form__section-head">{ Translations[ 'address-type-label' ] }</legend>
```
Abban, line 206: The language item for this header is quite long causing a line break on small screens. This will work better you remove `height: 18px;` from the `.form__section-head` in `DonationForm.pcss`

mobile/styles/layout/SelectGroup.pcss
```diff
@@ -5,23 +5,13 @@
...
-		margin-right: -4px;
-
-		&:after {
-			content: '';
-			display: table;
-			clear: both;
-		}
+		display: flex;
+		flex-wrap: wrap;
+		gap: 8px;
```
Abban, line 10: `gap` should work on mobile devices but be careful using it on a desktop banner: [see Can I Use](https://caniuse.com/?search=gap). A universal solution would be `margin -4px;` on the `.select-group`, and `padding: 4px;` on the `.select-group__option`. The amount buttons will also work at `width: 50%` then.

Abban (approved):
Looks good to me!

PR #787, opened 2021-12-09
Desktop-wpde-03

Abban:
Ticket: https://phabricator.wikimedia.org/T297291


moiikana (approved):

shared/components/ui/ProgressBar.jsx
```diff
@@ -107,9 +107,6 @@ export default class ProgressBar extends Component {
 		const Translations = context;
 		const getMillion = n => this.props.formatters.millionFormatter( n / 1000000 );
 		const getDaysLeft = daysLeft => {
-			if ( props.daysLeft > 14 ) {
```
moiikana, line 110: not so sure if it's fine to delete this in the shared progress bar, but maybe you've already talked about that

Abban: No we didn't talk about it at all! I think that we should be manually hiding the text with css anyway instead of having it automatically appear.

We actually should check the existing banners that are going to have that text appear at the weekend and make sure it's styled though!

moiikana (changes requested):
might be missing the footer

moiikana (approved):

PR #789, opened 2021-12-10
C21_WMDE_Mobile_Test_08

Abban:
Based on the VAR of Mobile 07

Var has the theme of the old red banner

Ticket: https://phabricator.wikimedia.org/T297380


moiikana (approved):

moiikana:
updated use of funds hint
https://phabricator.wikimedia.org/T297766

PR #948, opened 2022-06-24
Remove duplicate Keen Slider script

Abban:
Keen Slider version 6 isn't working,
so it should be locked at 5 for now.

Ticket: https://phabricator.wikimedia.org/T302063


wikipedia.de/mobile/components/FullpageBanner_var.jsx
```diff
@@ -19,13 +19,9 @@ export default class FullpageBanner extends Component {
...
 					Jetzt spenden
 				</div>
-				<Infobox
-					formatters={props.formatters}
-					campaignParameters={props.campaignParameters}
-					campaignProjection={props.campaignProjection}
-					bannerText={props.bannerText}
-					propsForText={ { overallImpressionCount: props.impressionCounts.getOverallCount() } }
-				/>
+				<div className="infobox">
```
gbirke, line 22: Shouldn't we take the opportunity and rename this?

gbirke (approved):

wikipedia.de/mobile/components/FullpageBanner_var.jsx
Abban: I'd rather make a separate PR. There's different implmentations of the element in different banners, so going through them all together as separate commits to a single PR makes sense to me.

moiikana:
how the branch currently looks to me in firefox:
slider is working but width is too high
Bildschirmfoto vom 2022-07-05 17-05-52

PR #949, opened 2022-06-27
C22_WMDE_Test_01

Abban:
Comparing the ctrl of the last 2021 banner which is delayed against a version that loads immediately

Ticket: https://phabricator.wikimedia.org/T310985


shared/skin_immediate/Skin.js
```diff
@@ -0,0 +1,60 @@
+'use strict';
+
+// eslint-disable-next-line no-unused-vars
+import $ from 'jquery';
+
+export default class Skin {
```
gbirke: The only difference in this poor man's abstract class is the missing `addSpace`. I'd prefer if the `skin_immediate` classes would all inherit from the original `Skin` class and have a different implementation of `addSpace`(e.g. calling `addSpaceImmediate` internally or having the presenter call `addSpaceImmediate`).

shared/skin_immediate/minerva.js
```diff
@@ -0,0 +1,51 @@
+import Skin from './Skin';
+import $ from 'jquery';
+
+// usually mobile skin
+
+export default class Minerva extends Skin {
```
gbirke: This is a 1:1 copy of the original class. I'd rather have it inherit from the original Minerva class with a a note to do a custom implementation if needed.

gbirke (approved):
I don't like the duplication of the skin adjusters, but would be ok to merge them for now.

shared/skin_immediate/minerva.js
Abban: The Minerva updates are in https://github.com/wmde/fundraising-banners/pull/953

PR #980, opened 2022-07-27
C22_WMDE_Test_03

moiikana:
- sentence of appreciation

https://phabricator.wikimedia.org/T313715


desktop/components/Banner_var.jsx
```diff
@@ -0,0 +1,264 @@
+import { Component, h, createRef } from 'preact';
```
gbirke: It's really unfortunate that we have to duplicate the whole Banner component just for changing the content. Maybe you can change the components to pass in the slides as a prop from the entry point, just like we do with the text. You can do that as a followup pull request

gbirke (approved):

PR #1003, opened 2022-08-18
Fix customamount field submit

moiikana:
- the validation should work not only on the submit button of the form
but also if you press [enter] while focussing the customAmount field.

- previously the validation got skipped when pressing [enter]
on the custom Amount input field (with an invalid value inside)

- the fix is now applied to all channels,
also the channels that use a 2step form

- also fixes wrong background on pad-EN ctrl banner in the customAmount field


Abban (approved):
We must try and work out a way to put the multi-step donation form into shared at some point. It'd be tricky though, because they don't all have the same functionality.

PR #1017, opened 2022-09-07
Component restructure

Abban:
TODO:

- [x] Use `components/Slider/Slider` (jsx and pcss)  in `MiniBanner`, removing the import of `Slider.pcss` in `styles.css`. This probably involves porting the classes
- [x] Copy `FollowupTransition` to `components/Transitions` folder (maybe adapting the classes). Just for consistencies sake.
- [x] Fix `gohan` theme for larger breakpoints (form buttons collapse into each other)
- [x] Discuss how we'll structure the "Banner Styles" (`mobile_rebuild/styles`) - does the split into `theme` and `layout` still make sense for the mobile banner? I'd find it easier if those files were unified again. If you decide to keep them split, please write reasoning/instructions in the Google Doc
- [x] Discuss file structure of `archive` folder (see my comment below).
- [x] Redo Test 05
  - [x] Copy `banners/desktop` to archive folder.
  - [x] Add new close button component and use in var, adapt campaign_config
- [x] Delete old `desktop` and `mobile` folders, rename `desktop_rebuild` and `mobile_rebuild`
- [x] merge this PR
- [ ] (Optional Cherry on Top) Document all the properties of `DonationForm` with `PropTypes` (see `TextHighlight` or `ProgressBar` as examples)


gbirke (approved):

gbirke:
I think we haven't discussed the concept and purpose of the `archive` folder enough, so I'm making this comment to start off the discussion:

Currently, it has two purposes:

1) Act as an easy starting point if the Fundraising team wants to revive an old banner (depending on the hierarchy, we can just copy the folder contents back into `banners/CHANNEL-NAME`)
2) A semi-easy way to look at old banners (just add an entry to campaign_info.toml). We should only have 1-2 old banners in there to avoid exploding build times. But it's an important feature to check if our changes would break old banners and to visually compare the "new" banner with the one that it's based on during development.
3) Reminder for us and the IDE to where components and style files are used. For Components we can use the "Navigate - Declaration or Usage" menu item, For styles that's not possible because of our heavy ampersand-nesting usage (we'll have to grep for file names in `archive` instead).

The folder structure is still open for discussion. The question is how you call them and how deep you nest them.

- With a flat folder structure that follows the campaign names (`archive/C22_WMDE_Test_04`, `archive/C22_WMDE_Test_Mobile_02`, etc), we have the benefit of easy archiving: Just copy the folder from `banner/CHANNEL-NAME` to `archive/BRANCH-NAME`. It will be sorted in a sensible way (year-platform-channel-number). The drawback is many folders in the archive: After the campaign there will be around 60 (based on counting the 2022 roadmap). I don't think it'll be a problem, the sorting makes it ok to navigate and find things by scrolling.
- A more nested folder structure, e.g `archive/desktop/2022/05` or `archive/2022/desktop/05` has the benefit of easy click navigation in the IDE file tree. But it has the drawback that we'll have to adapt all the js and css imports after copying it (and changing them back if we revive an old banner). We can automate this via npm scripts or shell tools (have an `archive` and an `unarchive` script that does the copying and modifies relative paths), but that adds complexity and possible brittleness.

I'm open for both of them and have a slight preference for the flat structure. Please discuss and choose whatever you like better (maybe also add a comment here). 

Future (ginger beer & cake) plans for the archive folder, to make it even more useful:
- usage graphs for easier checks what's still used after a cleanup. 
- Integrate navigating and viewing old banners in the archive into the dev environment overview, so we don't need to modify campaign_info.toml but still have ok build times.

moiikana:
on the archive folder structure:

- flat option would be easier to handle while archiving

- but the lookup of a design in a 60+ elements folder might be really too painful, so +1 for nested structure
  - I support the grouping by channel (`archive/desktop/2022/05`) because it reflects better what we need in our daily workflow (looking up how something looked on a specific channel before)
   Cleaning away old unused banner designs (from 2-3 years ago) from the archive folder happens more rarely (so the year does not have to be a main folder group)

PR #1034, opened 2022-09-22
C22_WMDE_Mobile_Test_05 (button layout change in minibanner)

moiikana:
- variant has 2 buttons on the mini banner

https://phabricator.wikimedia.org/T318073


Abban (changes requested):
Looks good, but I noticed that the button text on the var minibanner button text has a line-break at 320px width. We should add a breakpoint to shrink the text a little before that happens. Edit: There's also a line break at 430px

moiikana:
> Looks good, but I noticed that the button text on the var minibanner button text has a line-break at 320px width. We should add a breakpoint to shrink the text a little before that happens. Edit: There's also a line break at 430px

I could not replicate any text line-break behaviour on the orange button in the var minibanner >=320px.
I tested in firefox dev tools and chromium dev tools

Update: was because of the Lato font,
added a CSS fix to prevent the line from breaking for now, 
might be more changes necessary in the future for different typefaces and their sizes.

Abban (approved):

PR #1050, opened 2022-10-04
Disable certain interval and payment combinations

gbirke:
Check for certain combinations of payment interval and method and create a list of disabled values for the complementary value.

Reorder imports to put components and hooks together.

Fix linter warning in DonationFormPreselectableAmount.


Abban (approved):
This is much better than the old way of doing it.

PR #1053, opened 2022-10-06
C22_WMDE_Test_EN_01

Abban:
* Used desktop as the base and added the progress bar and English localisations.
* Update config and text
* Use new beg form for var and add styles to handle it to the Goku theme

Ticket: https://phabricator.wikimedia.org/T318651


gbirke (approved):
The new visitors-vs-donors sentence shows "only 0 people have donated" for me, I don't know if that's working correctly or just a glitch in my test setup. 
In VAR, the sentence is also not highlighted, I think we need to make it consistent.

components/UseOfFunds/FundsContent.jsx
```diff
@@ -68,6 +68,7 @@ export default function FundsContent( props ) {
 		<div className="banner_model__section use_of_funds__section--call_to_action">
 			<button className="use_of_funds__button" onClick={onCallToAction}>{content.callToAction}</button>
 		</div>
+		{ props.figuresAreProvisional && ( <div style="text-align: left; font-size: small; padding-bottom: 16px;">*) vorläufig</div> ) }
```
gbirke:
```suggestion
		{ props.figuresAreProvisional && ( <div style="text-align: left; font-size: small; padding-bottom: 16px;">{ content.provisional }</div> ) }
```

PR #1054, opened 2022-10-06
C22_WMDE_ipad_Test_01

Abban:
* The base is the CTRL of last years ipad-de-02.
* Added a new theme for the red/black/white banners

Ticket: https://phabricator.wikimedia.org/T318647


campaign_info.toml
```diff
@@ -82,20 +82,20 @@ orientation = [ "portrait", "landscape" ]
...
 [pad]
-campaign_tracking = "pad02-ba-211119"
+campaign_tracking = "pad01-ba-221103"
 preview_link = "/wiki/Wikipedia:Hauptseite?devbanner={{banner}}&banner=B17WMDE_webpack_prototype"
 preview_url = 'https://de.wikipedia.org/wiki/Wikipedia:Hauptseite?banner={{PLACEHOLDER}}'
 wrapper_template = "wikipedia_org"
 
 [pad.banners.ctrl]
 filename = "./banners/pad/banner_ctrl.js"
-pagename = "B21_WMDE_ipad_02_211119_ctrl"
```
moiikana, line 92: there is a pre-shutterbug era leftover:
the pagename should not contain the date, it has to match up with the preview link on Central Notice, that means it has to be 
B22_WMDE_ipad_Test_01_ctrl
and
B22_WMDE_ipad_Test_01_var
in order to work with our `banner-screenshots` code and the requests to testingbot.

same with the EN PAD test and all other banners that we haven't change a lot during the past months

Abban: Probably why shutterbug failed

gbirke: I've updated https://github.com/wmde/fundraising-infrastructure/wiki/How-to-manage-banners-on-de.wikipedia.org with our current banner naming schema. Previously it still recommended putting the start date into the banner name, but we haven't been doing that since the 2019 campaign.

gbirke (approved):
Code is fine, disguised radio buttons are a usability flaw IMHO (should use the mobile one for ipad as well), but that's for the PM to decide.

banners/pad/components/Banner_var.jsx
```diff
@@ -1,47 +1,66 @@
...
+import ButtonClose from '../../../components/ButtonClose/ButtonClose';
+
+const PENDING = 0;
+const VISIBLE = 1;
+const CLOSED = 2;
 
 const SLIDESHOW_START_DELAY = 2000;
 const SLIDESHOW_SLIDE_INTERVAL = 10000;
 
-const BannerVisibilityState = Object.freeze( {
```
moiikana, line 22: why did the BannerVisbilityState object/enum get turned back into single constants again?

Abban: Copy/paste error I think, I probably copied over the component from the English iPad or WPDE test.

PR #1056, opened 2022-10-07
Add Multistep Beg Yearly form component

Abban:
This component is to be used in banners that implement this year's 2 step process

### DIFF
--- a/components/DonationForm/BegYearlyRecurringDonationForm.jsx
+++ b/components/DonationForm/BegYearlyRecurringDonationForm.jsx
@@ -49,6 +49,9 @@ export default function BegYearlyRecurringDonationForm( props ) {
 	);
 
 	const onSubmitStep1 = e => {
+		if ( customAmount ) {
+			validateCustomAmount( customAmount );
+		}
 		if ( [
 			[ intervalValidity, setIntervalValidity ],
 			[ amountValidity, setAmountValidity ],
@@ -201,7 +204,7 @@ export default function BegYearlyRecurringDonationForm( props ) {
 							fieldname="select-amount"
 							value={ customAmount }
 							selectedAmount={ selectedAmount }
-							onInput={ e => { updateCustomAmount( e.target.value ); validateCustomAmount( e.target.value ); } }
+							onInput={ e => updateCustomAmount( e.target.value ) }
 							onBlur={ e => validateCustomAmount( e.target.value ) }
 							placeholder={ props.customAmountPlaceholder }
 							language={
--- a/components/DonationForm/BegYearlyRecurringDonationForm.jsx
+++ b/components/DonationForm/BegYearlyRecurringDonationForm.jsx
@@ -16,7 +16,6 @@ import useFormAction from './hooks/use_form_action';
 import { amountMessage, validateRequired } from './utils';
 import { Intervals, PaymentMethods } from '../../shared/components/ui/form/FormItemsBuilder';
 import classNames from 'classnames';
-import ChevronLeftIcon from '../Icons/ChevronLeftIcon';
 
 const formSteps = Object.freeze( {
 	ONE: Symbol( 'one' ),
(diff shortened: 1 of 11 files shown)


moiikana:
might already make sense at this point to have a `components/DonationForm/Forms/` folder to put in the different donation form implementations

components/DonationForm/BegYearlyRecurringDonationForm.jsx
```diff
@@ -0,0 +1,292 @@
...
+const formSteps = Object.freeze( {
+	ONE: Symbol( 'one' ),
+	TWO: Symbol( 'two' )
+} );
+
+export default function BegYearlyRecurringDonationForm( props ) {
+	const Translations = useContext( TranslationContext );
+	const [ paymentInterval, setInterval, intervalValidity, setIntervalValidity ] = useInterval( null );
+	const [ paymentMethod, setPaymentMethod, paymentMethodValidity, setPaymentMethodValidity ] = usePaymentMethod( null );
+	const [ alternative, setAlternative, alternativeValidity, setAlternativeValidity ] = useAlternate( null );
```
moiikana: maybe it is possible to describe / rename that `Alternative` hook here, or add an explanatory comment in the hook js file to describe what it is doing, because `Alternative` leaves a lot of question marks in my head when browsing the code

moiikana (changes requested):

components/DonationForm/hooks/use_alternative.js
```diff
@@ -0,0 +1,16 @@
+import { useState } from 'preact/hooks';
+import { UNSET, VALID } from './validation_states';
+
+const alternativeIsValid = alternative => alternative === null ? UNSET : VALID;
+
+export const Alternatives = { NO: 'NO', YES: 'YES' };
+
+export default function useAddressType( initial ) {
```
moiikana: function title has to change, it's still useAddressType

moiikana (changes requested):

components/DonationForm/BegYearlyRecurringDonationForm.jsx
Abban: The field is called alternative, though. The second step is basically "hey we got your choices, would you like to do this instead?"

moiikana: I tried to come up with an alternative to Alternative. I think in my mind it would be more like `SuggestedValue` or something like that. Or `SuggestedAlternative`, even though it should of course be kept short.
`Alternative` feels a slight bit too ambiguous as a name/identifier in code for me.
I would like to have another opinion on it by @gbirke and/or @TanujaDoriya, then I'll approve

gbirke: I also find the name "Alternative" too ambiguous. I like the hook for a "binary state" (that in reality is a trinary state because it can also be  "Unset") - we can reuse that for other options. Suggested names (for the state, use `use_`, `set` `validity` prefixes/suffixes as needed): `binaryState`, `yesNoState`, `twoOptionsState`, `twoAlternativesState`, etc

gbirke: When using the hook, we should also name the variables after the domain thing they represent, e.g. 
```js
const [ hasUpgradedToRecurring, upgradeToRecurring, upgradeToRecurringIsValid, setupgradeToRecurringValidity ] = useYesNoState( null );
```

Abban: I don't get why this is an issue now after being used in multiple banners in the last campaign (and maybe the one before)? `yesNoState` means even less than `alternative`

Abban: This branch is rebased into multiple current banners now so this change will have to be made in another PR. I made a ticket for it. https://phabricator.wikimedia.org/T320467

components/DonationForm/BegYearlyRecurringDonationForm.jsx
```diff
@@ -0,0 +1,246 @@
...
+				</fieldset>
+
+				<div className="wmde-banner-form-button-container">
+					<button className="wmde-banner-form-button" type="submit">
+						{ step1ButtonText }
+					</button>
+				</div>
+			</div>
+
+			<FormStep2
```
gbirke, line 241: Anticipating the mobile banner with a different implementation/ markup of the choices (buttons instead of radio buttons), you could make this a prop and pass it in from the banner component. But maybe that's premature, so I just want to point that out.

Abban:
This can be merged now if someone can re-re-review it

gbirke (approved):

PR #1058, opened 2022-10-10
C22_WMDE_Test_08

Abban:
The base banner is based off the `CTRL` of test 7.

The VAR implements the new `BegYearlyRecurringDonationForm`.

Ticket: https://phabricator.wikimedia.org/T318661


gbirke (approved):
The form items of VAR have slightly different whitespace than the ones in CTRL, but it doesn't affect banner height, so I guess it's fine

PR #1065, opened 2022-10-10
Add current slide class to slider

Abban:
This will add a class to the current slider slide which allows us to target text animations in slides

This is used in https://github.com/wmde/fundraising-banners/pull/1054

### DIFF
--- a/components/Slider/Slider.jsx
+++ b/components/Slider/Slider.jsx
@@ -87,7 +87,10 @@ export default function Slider( props ) {
 			<div className="wmde-banner-navigation-wrapper">
 				<div ref={ sliderRef } className="keen-slider wmde-banner-slider" onMouseDown={ stopAutoplay } onTouchStart={ stopAutoplay }>
 					{ slides.map( ( slide, idx ) => (
-						<div key={ idx } className="keen-slider__slide wmde-banner-slide">
+						<div key={ idx } className={ classNames(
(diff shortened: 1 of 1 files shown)


components/Slider/Slider.jsx
```diff
@@ -87,7 +87,10 @@ export default function Slider( props ) {
 			<div className="wmde-banner-navigation-wrapper">
 				<div ref={ sliderRef } className="keen-slider wmde-banner-slider" onMouseDown={ stopAutoplay } onTouchStart={ stopAutoplay }>
 					{ slides.map( ( slide, idx ) => (
-						<div key={ idx } className="keen-slider__slide wmde-banner-slide">
+						<div key={ idx } className={ classNames(
+							'keen-slider__slide wmde-banner-slide',
+							{ 'wmde-banner-slide--current': slider && slider.details().absoluteSlide === idx }
```
gbirke: The `details()` method is probably just a getter, but having a method called multiple times in a loop can be a performance bottleneck. Can you assign the value to a `const` (e.g. `currentSlideIndex`) outside of the `return` section or will that break because of the the way reactivity works?

gbirke (approved):

components/Slider/Slider.jsx
Abban: There was actually already a reactive item for it.. so I swapped out the variable

gbirke (approved):

PR #1066, opened 2022-10-10
Add Mr Satan theme

Abban:
This is used on the iPad and WPDE desktop banners


gbirke (changes requested):
I find it hard to review without an example

Abban:
Ah right, I was shooting through all these PRs and didn't think about that. It's being used in https://github.com/wmde/fundraising-banners/pull/1054

Abban:
The styles for the multistep form might change when implemented into the mobile banners whiuch will require them to be updated here, so don't merge this yet

gbirke (approved):

PR #1069, opened 2022-10-11
C22_WPDE_Test_01

Abban:
Convert last year's test 03 to the new structure

Ticket: https://phabricator.wikimedia.org/T318657


gbirke (changes requested):
The VAR banner is slightly bigger than the CTRL, with a lot of whitespace on the 2nd page

Abban:
>The VAR banner is slightly bigger than the CTRL, with a lot of whitespace on the 2nd page

Good catch, I had a hardcoded padding value on the form. Fixed by https://github.com/wmde/fundraising-banners/pull/1091

Sperling-0 (commented):
"The VAR banner is slightly bigger than the CTRL, with a lot of whitespace on the 2nd page" : This issue persists on my Chromium browser. However no such issue when using Firefox.

banners/wpde_desktop/styles/Banner.pcss
```diff
@@ -0,0 +1,61 @@
...
+	&-content {
+		display: flex;
+		flex-direction: row;
+		padding: 8px $banner-padding 0;
+	}
+
+	&-column-left {
+		display: flex;
+		align-items: center;
+		border: 8px solid $color-primary;
```
moiikana, line 39: the styling elements here would go to the themes, right?

moiikana:
the `variant-wpde-desktop.pcss` will become a theme next? or what is the purpose of it in the theme?

gbirke (approved):

Abban:
> the variant-wpde-desktop.pcss will become a theme next? or what is the purpose of it in the theme?

There was a few little differences between the WPDE banner and the iPad one that use the same theme. An entire variant for those things felt like overkill so I stuck it into that file

PR #1076, opened 2022-10-13
C22_WMDE_Mobile_EN_Test_01

Abban:
Add new theme: Videl

Ticket: https://phabricator.wikimedia.org/T318655


gbirke (approved):

components/UseOfFunds/FundsContent.jsx
```diff
@@ -68,6 +68,7 @@ export default function FundsContent( props ) {
 		<div className="banner_model__section use_of_funds__section--call_to_action">
 			<button className="use_of_funds__button" onClick={onCallToAction}>{content.callToAction}</button>
 		</div>
+		{ props.figuresAreProvisional && ( <div style="text-align: left; font-size: small; padding-bottom: 16px;">*) vorläufig</div> ) }
```
gbirke: This should be in English

banners/mobile_english/translations.js
```diff
@@ -1,9 +1,8 @@
 const Translations = {
-	'intervals-header': 'I will donate:',
```
moiikana, line 2: probably better to hide the headers with css collectively than emptying the translation keys

Abban: I looked up what's going on. These translation items go into the form `legend` elements and while the form doesn't have text it styles the empty legends as fieldgroup separators. It's not very semantic, so I added the text, hid the legend elements with CSS and added the visual separator to the fieldgroups instead.

PR #1078, opened 2022-10-17
C22_WMDE_ipad_en_Test_01

Abban:
VAR banner uses the new Soft Close functionality

Ticket: https://phabricator.wikimedia.org/T320466


components/SoftClose/SoftClose.jsx
```diff
@@ -0,0 +1,64 @@
...
+	render( props, state, context ) {
+		const Translations = context;
+		return ( <div className="wmde-banner-soft-close">
+			<div className="wmde-banner-soft-close-countdown-bar">
+				<div className="wmde-banner-soft-close-countdown-bar-fill"></div>
+			</div>
+			<div className="wmde-banner-soft-close-columns">
+
+				<div className="wmde-banner-soft-close-column">
+					<span className="wmde-banner-soft-close-prompt">{ Translations[ 'soft-close-headline' ] }</span>
```
gbirke: Could we change the translation keys to match the class, `soft-close-prompt`?

components/SoftClose/SoftClose.jsx
```diff
@@ -0,0 +1,64 @@
...
+						{ Translations[ 'soft-close-button-1' ] }
+					</button>
+					<button
+						className="wmde-banner-soft-close-button"
+						onClick={ props.onCloseBanner }>
+						{ Translations[ 'soft-close-button-2' ] }
+					</button>
+				</div>
+
+				<div className="wmde-banner-soft-close-column wmde-banner-soft-close-countdown-text" dangerouslySetInnerHTML={ {
```
gbirke: Could we change the translation keys to match the class, `soft-close-countdown-text`?

gbirke (commented):
When clicking "yes" in the soft-close component the big banner shows again until the timeout runs out. Could not figure out why yet ...

Abban:
>When clicking "yes" in the soft-close component the big banner shows again until the timeout runs out. Could not figure out why yet ...

I'd forgotten to rename some state variables 🤦

moiikana:
what is the difference between /shared/components/ and /components/ ? the imports seem to come variously from both sources

gbirke:
> what is the difference between /shared/components/ and /components/ ? the imports seem to come variously from both sources

There are just 2 things that come from `/shared/components`: TranslationContext and everything related to use of funds. We can create copies of those components in `/components/` and update the code in the next round of banners. AFAIR for use of funds we did not want to duplicate the necessary changes until all banners were ready and TranslationContext just slipped though.

If you see any other components imported from `/shared/components`, please flag those. We should be able to get rid of them soon.

moiikana:
> > what is the difference between /shared/components/ and /components/ ? the imports seem to come variously from both sources
> 
> There are just 2 things that come from `/shared/components`: TranslationContext and everything related to use of funds. We can create copies of those components in `/components/` and update the code in the next round of banners. AFAIR for use of funds we did not want to duplicate the necessary changes until all banners were ready and TranslationContext just slipped though.
> 
> If you see any other components imported from `/shared/components`, please flag those. We should be able to get rid of them soon.

Would be nice to have the /shared/components removed or renamed, as it is an easy ground for confusion for people like me who haven't work with it a lot, if this section is not cleaned up.
As there are still different PRs open on this, I'm not sure if / where the cleanup happens (or not)

Abban:
I'll delete the shared/components folder as part of the thank you banner work as that's the last one to use components from there.

moiikana (approved):

PR #1084, opened 2022-10-18
C22_WMDE_Mobile_Test_05_2

Abban:
Update configuration for C22_WMDE_Mobile_Test_05_2. This is so we can tag the specific banner release that was run after it was fixed.

Ticket: https://phabricator.wikimedia.org/T320601

### DIFF
--- a/campaign_info.toml
+++ b/campaign_info.toml
@@ -38,20 +38,20 @@ tracking = "org-06-220916-ctrl"
 
 # Mobile Campaign
 [mobile]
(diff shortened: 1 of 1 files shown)


gbirke (changes requested):
You should also change the `pagename` to use the new banners with the changed tracking

Abban:
> You should also change the `pagename` to use the new banners with the changed tracking

Done.

gbirke (approved):

PR #1086, opened 2022-10-18
Theme animated text

Abban:
This adds an explicit file for the animated banner text into the Goku and Mr Satan themes so all the animated styles are now in one place.


gbirke (approved):

Abban:
You can see the Gohan text working in https://github.com/wmde/fundraising-banners/pull/1073

PR #1092, opened 2022-10-20
Hide banner when minerva search is clicked

Abban:
The banner wasn't being hidden when the user clicked the magnifying glass icon


gbirke (approved):
Tested, works, great job!

PR #1093, opened 2022-10-21
Fix layout and text issues in UoF accordion

Abban:


gbirke (approved):

gbirke:
You might want to correct the misspelling of "accordion" in the commit msg. I corrected it for the PR

PR #1104, opened 2022-10-26
Validate custom amount on submit only

Abban:
Validation was added to the custom amount
onValueChanged event in order to stop people
getting past validation by pressing enter
on their keyboards. This prevented people
being able to donate decimals, however.

This fix moves the custom amount validation
to the onSubmit method instead

You can see this working in https://www.wikipedia.de/?banner=B22WPDE_mob01_ctrl


gbirke (commented):
Thanks for putting a description in the PR, this has really helped to figuring out what this was about.

gbirke (approved):

PR #1108, opened 2022-10-28
Updates for Progress Bar

Abban:
This contains updates for the various versions of the progress bar.

### DIFF
--- a/components/ProgressBar/ProgressBar.jsx
+++ b/components/ProgressBar/ProgressBar.jsx
@@ -102,8 +102,7 @@ export default class ProgressBar extends Component {
 	}
 
 	calculateWidth() {
-		const max = Math.max( ( this.props.donationAmount * 100 ) / this.props.goalDonationSum, 20 );
-		return Math.min( max, 100 );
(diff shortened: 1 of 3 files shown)


components/ProgressBar/ProgressBar.jsx
```diff
@@ -102,7 +102,7 @@ export default class ProgressBar extends Component {
 	}
 
 	calculateWidth() {
-		const max = Math.max( ( this.props.donationAmount * 100 ) / this.props.goalDonationSum, 20 );
```
gbirke, line 105: This would mean that the bar is always filled by 20% even at the start of the campaign when the value is realistically at 10%. I think that's a bit much, let's ask the PM about that ...

Are there instances where that 20% is needed to fit in the text because the `min-width` doesn't take? Then I might be ok with this, but still would like to investigate why `min-width` doesn't work ...

Abban: Yeah, it is for fitting the text. There was a reason for it but after a whole 3 days I can't remember what it was...

Sperling-0 (approved):

PR #1110, opened 2022-10-31
Mr Satan Tweaks

Abban:
Tidy up a little weirdness in the slider left over from when the banners used to break into a single column on small sizes.

Fix an error with IE11 and validation error icons.

### DIFF
--- a/themes/mr-satan/DonationForm/SelectGroup.pcss
+++ b/themes/mr-satan/DonationForm/SelectGroup.pcss
@@ -25,7 +25,8 @@
 				padding-left: 20px;
 
 				&:before {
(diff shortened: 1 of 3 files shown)


Sperling-0 (approved):

moiikana:
what was the reason that the icon moved out of the variables file into the code? Did IE11 require the move? if this is not finished yet ignore this comment

Abban:
>what was the reason that the icon moved out of the variables file into the code? Did IE11 require the move?

IE11 didn't like it in the WPDE desktop banner. Just that banner, it was fine with the exact same code elsewhere.

PR #1118, opened 2022-11-01
Allow transitioning skin space removal

Abban:
You can see this working in https://github.com/wmde/fundraising-banners/pull/1106


gbirke (approved):
This is a BC breaking change to the BannerPresenter interface. You could have created a `removeSpaceWithTransiation` instead, but I like your approach for consistency.
I grepped through the code and the only affected banner is the old thank you banner. Could you adapt that (in this PR or a followup), so we don't forget to do the change and waste time on debugging when we finally get to the TY banners?

Abban:
I think that the code in that file isn't being used anywhere. Maybe it's from the old banner environment?

Abban:
I'm going to convert the thank you banner to the new structure next week. I'll delete that file when I do it.

PR #1122, opened 2022-11-09
C22_WMDE_Mobile_Test_08

Abban:
Var has a copy change on the modal.

Ticket: https://phabricator.wikimedia.org/T322730


archive/mobile/C22_WMDE_Mobile_Test_07/styles/MiniBanner.pcss
```diff
@@ -0,0 +1,128 @@
...
+		}
+
+		&-button {
+			width: 100%;
+			height: 40px;
+			line-height: 40px;
+			border-radius: 20px;
+			font-weight: bold;
+			color: $color-white;
+			background: $color-secondary;
```
moiikana, line 104: some color definitions here belong to the themes, I think

Abban: So far the convention I was using it to keep the css that styles the elements in the MiniBanner.jsx and FullBanner.jsx in the banner.

But you're right, only the layout needs to stay so a lot of that can go into the theme. 🤔

gbirke: As long as these styles belong to elements that are in `Minibanner.jsx`, I would rather not move them to themes, because themes are mostly for the extracted elements of banners and should not know about the banner elements. I've created a ticket to extract more small elements (close button, submit button) out of the banners and into shared components: https://phabricator.wikimedia.org/T323100

But that is not in the scope of this ticket and a review criterion for new banners

If you see style "overrides" for elements that are already extracted, then sound the alarm, because that's a habit that will degrade our pretty structure a little bit, each time we do it, eventually leading to the mess we tried to escape from. But as long as the banner-specific CSS targets banner-specific markup, it's fine.

PR #1123, opened 2022-11-09
C22_WMDE_Test_EN_02

Abban:
VAR banner uses the new soft close functionality

Ticket: https://phabricator.wikimedia.org/T322731


components/SoftClose/SoftClose.pcss
```diff
@@ -20,6 +20,7 @@
 		}
 		.wmde-banner-soft-close-countdown-bar-fill {
 			animation: wmde-countdown-bar 15s linear;
+			animation-fill-mode: forwards;
```
moiikana, line 23: what is the purpose of this line?

Abban: It stops the animation at the last frame rather than resetting to the first one

Sperling-0 (approved):

PR #1125, opened 2022-11-14
C22_WMDE_Test_12

Abban:
VAR banner implements new Subscription Form

Ticket: https://phabricator.wikimedia.org/T323021


banners/desktop/banner_var.js
```diff
@@ -50,7 +51,7 @@ bannerPresenter.present(
 		footer: Footer,
 		bannerText: BannerText,
 		slides: Slides,
-		translations: Object.assign( Translations, LocalTranslations, TranslationsSoftClose ),
+		translations: Object.assign( Translations, LocalTranslations, TranslationsSoftClose, TranslationsSubscriptionForm ),
```
gbirke: If softclose is now a permanently accepted feature in banners, we should merge it into local translations

components/SubscriptionForm/SubscriptionForm.jsx
```diff
@@ -0,0 +1,112 @@
...
+		} );
+	}
+
+	setConsentGiven( e ) {
+		this.setState( { consentGiven: e.target.checked } );
+	}
+
+	emailIsValid( email ) {
+		return String( email )
+			.toLowerCase()
```
gbirke, line 50: Your regex does not distinguish between cases, to the `.toLowerCase()` is unneccessary IMHO. However, the user could accidentally type a space at the bginning or end of the email which would lead to hard-to-discover validation errors with your regex, so I recommend adding a `.trim()` (and testing if our transpiler replaces `trim` with an IE-compatible version and if not, stripping leading and training whitespace with [a manual polyfill](https://gist.github.com/arturotena/22158dcabe386bbfc536)

gbirke (changes requested):

banners/desktop/banner_var.js
Abban: I quite like them separated like this as it's easier to know what's what. Maybe keep them in separate files and move them into a subfolder?

components/SubscriptionForm/SubscriptionForm.jsx
Abban: `trim()` seems to work fine with IE11

banners/desktop/banner_var.js
gbirke: Yeah, maybe having default translations for features like soft close or the 2nd steps of the donation form is a good idea. Then they are cleanly separated and can still be overridden by LocalTranslations (which should come at the very end of that list)

gbirke (approved):

banners/desktop/banner_var.js
Abban: I'm gonna move the defaults in with the component

PR #1131, opened 2022-11-21
Fix main and rename mobile form 2

gbirke:


moiikana (approved):

moiikana:
I think you have to do the renaming from the second commit in the archived banners as well. When we restore them we would want them to work with our current components.

moiikana (changes requested):
probably necessary to also change the imports of `begyearlyrecurringdonationformstep2mobile` in the archive, that way we could restore an archived banner easier with our current components

gbirke:
I have split the main fix into #1132 so I can do the rename thing properly, whithout descending into rebase hell

gbirke:
I've split this and have a new PR #1134 which addresses your comment

PR #1138, opened 2022-11-25
Update only the content dependency

moiikana:
we often need to keep the content up to date for the banners, but other dependencies should not always be updated to not break some functionality during the campaign for example

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,9 @@ ci: test lint ## Run continuous integration tasks (linting, testing)
 clean: ## Clean up dist directory
 	docker-compose run js-build npm run clean
 
+update-content:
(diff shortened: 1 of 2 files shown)


gbirke (changes requested):
Just want to point out that this is the first change to the Makefile that doesn't use Docker. I like the consistency with the fundraising app, so I'm not opposed to putting it in the Makefile, but for more consistency inside this repo we should prefix it with a docker-compose command (`docker-compose run js-build`). To make the command itself shorter, you could add a `update-content` script to `package.json` (calling `npm update `), so people who aren't using docker can run `npm run update-content`

Abban (approved):

PR #1139, opened 2022-11-25
Allow delay handling via query params

moiikana:
- code checks for "&devMode" parameter, else 2 fallback options will be applied,
the last option is the delay on production

https://phabricator.wikimedia.org/T323691

### DIFF
--- a/shared/banner_delay.js
+++ b/shared/banner_delay.js
@@ -1,21 +1,7 @@
-export default class BannerDelay {
-
-	constructor( bannerContainer ) {
-		this.bannerContainer = bannerContainer;
+export default function getBannerDelay( delayValue ) {
+	const parameters = window.location.search;
+	if ( parameters.match( /&devMode/ ) ) {
+		return 1;
 	}
-
-	getBannerDelay( delayValue ) {
-		const url = window.location.href;
-		const parameters = url.split( '?' )[ 1 ];
-
-		if ( parameters.includes( '&noDelay' ) ) {
-			return 1;
-		}
-
-		if ( this.bannerContainer.dataset.delay ) {
-			return this.bannerContainer.dataset.delay;
-		}
-
-		return delayValue;
-	}
+	return delayValue;
 }
--- a/shared/banner_delay.js
+++ b/shared/banner_delay.js
@@ -1,0 +1,21 @@
+export default class BannerDelay {
+
+	constructor( bannerContainer ) {
+		this.bannerContainer = bannerContainer;
+	}
+
+	getBannerDelay( delayValue ) {
+		const url = window.location.href;
+		const parameters = url.split( '?' )[ 1 ];
+
+		if ( parameters.includes( '&noDelay' ) ) {
+			return 1;
+		}
+
+		if ( this.bannerContainer.dataset.delay ) {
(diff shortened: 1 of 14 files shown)


shared/banner_delay.js
```diff
@@ -0,0 +1,21 @@
+export default class BannerDelay {
+
+	constructor( bannerContainer ) {
+		this.bannerContainer = bannerContainer;
+	}
+
+	getBannerDelay( delayValue ) {
+		const url = window.location.href;
+		const parameters = url.split( '?' )[ 1 ];
```
gbirke:
```suggestion
		const parameters = window.location.search;
```

shared/banner_delay.js
```diff
@@ -0,0 +1,21 @@
...
+
+		if ( parameters.includes( '&noDelay' ) ) {
```
gbirke:
```suggestion
		if ( parameters.match( /&noDelay/ ) ) {
```

`string.includes` is a modern feature where either our transpiler would add polyfill code for IE or IE would break.

gbirke (changes requested):
Thank you for doing this, it works 100% like specified in the ticket.

Unfortunately I did not update the ticket after we talked in the daily about adding a `devMode` parameter (that would also set the event fire rate to 100%). So that would be one small change - to rename `noDelay` to `devMode`.

If you want to do further cleanup, you could make `getBannerDelay` a function, dropping the dependency on `bannerContainer`. The only place where that is used is for the dev banner, `B17WMDE_webpack_prototype`. We could replace that mechanism with adding `&devMode` to all URLs in `campaign_info.toml`. I know it's a change that touches all banners again and will do it myself if you don't want to do it.

moiikana:
@gbirke 
>The only place where that is used is for the dev banner, B17WMDE_webpack_prototype. We could replace that mechanism with adding &devMode to all URLs in campaign_info.toml.

I can add `&devMode` at the end of every preview_link in the campaign_info.toml, yes,
but I think `this.bannerContainer.dataset.delay` was taking care of that local banner development delay (1) anyway.

The purpose of this PR is also to make something like 
`https://de.wikipedia.org/?banner=B22_WMDE_Test_04_var&devMode`
available for Kai for acceptance testing, right?

I added a new commit with changes, please have a look

shared/banner_delay.js
```diff
@@ -0,0 +1,13 @@
+export default function getBannerDelay( bannerContainer, delayValue ) {
+	const parameters = window.location.search;
+
+	if ( parameters.match( /&noDelay/ ) ) {
```
gbirke:
```suggestion
	if ( parameters.match( /&devMode/ ) ) {
```

shared/banner_delay.js
```diff
@@ -0,0 +1,13 @@
...
+		return 1;
+	}
+
+	if ( bannerContainer.dataset.delay ) {
+		return bannerContainer.dataset.delay;
+	}
```
gbirke: With `devMode` in the campaign_info.toml, you can now completely drop `bannerContainer` as a parameter and in this check.

gbirke (approved):
I was thinking of dropping

gbirke (approved):

PR #1149, opened 2022-12-05
C22_WMDE_Mobile_EN_Test_02

moiikana:
https://phabricator.wikimedia.org/T324444


banners/mobile_english/styles/theme_overrides_ctrl.pcss
```diff
@@ -0,0 +1,3 @@
+.wmde-banner-navigation-wrapper {
+	padding-bottom: 57px !important;
+}
```
gbirke: I figured out how to override the theme file without using `!important`

1. Use the same nesting rules as in the original files. Apparently, PostCSS puts the non-nested classes at the top and then adds the concatened classes from nested rules after, regardless of the order in the file.
2. Use the media query, it adds specificity and overrides previous rules

```suggestion
.wmde-banner {
   &-navigation-wrapper {
	padding-bottom: 57px;
	@media ( min-width: $breakpoint-phone-s ) {
			padding-bottom: 57px;
	}
   }
}
```

banners/mobile_english/styles/theme_overrides_var.pcss
```diff
@@ -0,0 +1,7 @@
+.wmde-banner-slider-pagination {
+	padding-top: 10px !important;
+}
+
+.wmde-banner-navigation-wrapper {
+	padding-bottom: 55px !important;
+}
```
gbirke, line 12: Please change these rules as well

gbirke (changes requested):

moiikana:
rebased main

gbirke (approved):

PR #1157, opened 2022-12-09
C22_WMDE_ipad_en_Test_02

Abban:
* Updates text in both banners
* Introduces new currentDate property in dynamic text
* Both banners use the 3 step form
* Add soft close to VAR

Ticket: https://phabricator.wikimedia.org/T324821


banners/pad_english/styles/Banner.pcss
```diff
@@ -35,14 +35,22 @@
...
 		border: 5px solid $color-primary;
 		border-radius: 9px;
 		background-color: $color-white-off;
-		padding: 0 0 80px;
 		position: relative;
 		margin-right: $banner-padding;
 		overflow-y: hidden;
 
+		min-height: 345px;
+		@media ( min-width: 1114px ) {
```
moiikana, line 50: might this be `$breakpoint-m` ?

moiikana (approved):

banners/pad_english/styles/Banner.pcss
Abban: No, as it's to force the height to match the last banner and 1114px is where the text in that one wraps.

PR #1160, opened 2022-12-12
Only show banners on article namespace

Abban:
Ticket: https://phabricator.wikimedia.org/T324770

### DIFF
--- a/campaign_info.toml
+++ b/campaign_info.toml
@@ -54,7 +54,7 @@ orientation = [ "portrait", "landscape"]
 name = "C22_WMDE_Mobile_EN_Test_02"
 campaign_tracking = "mob_en02-ba-221206"
 preview_link = "/en-mobile/wiki/Main_Page?devbanner={{banner}}&useskin=minerva&banner=B17WMDE_webpack_prototype&devMode"
-preview_url = 'https://en.m.wikipedia.org/wiki/Wikipedia:Hauptseite?banner={{PLACEHOLDER}}'
+preview_url = 'https://en.m.wikipedia.org/wiki/Main_Page?banner={{PLACEHOLDER}}'
 wrapper_template = "wikipedia_org"
 
 [mobile_english.banners.ctrl]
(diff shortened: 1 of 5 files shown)


shared/banner_presenter.js
```diff
@@ -136,6 +136,16 @@ export default class BannerPresenter {
 			this.trackingData.viewportDimensionsTrackRatio
 		);
 
+		if ( onMediaWiki() && mw.config.values.wgNamespaceNumber !== 0 && mw.config.values.wgNamespaceNumber !== 4 ) {
```
gbirke: Please move this condition to a function like `isArticlePage` or `isWikipediaArticlePage`. Those namespaces are just an implementation detail

campaign_info.toml
```diff
@@ -53,8 +53,8 @@ orientation = [ "portrait", "landscape"]
 [mobile_english]
 name = "C22_WMDE_Mobile_EN_Test_01"
 campaign_tracking = "mob_en01-ba-221103"
-preview_link = "/en-mobile/wiki/Main_Page?devbanner={{banner}}&useskin=minerva&banner=B17WMDE_webpack_prototype&devMode"
-preview_url = 'https://en.m.wikipedia.org/wiki/Wikipedia:Hauptseite?banner={{PLACEHOLDER}}'
+preview_link = "/en-mobile/wiki/German_Wikipedia?devbanner={{banner}}&useskin=minerva&banner=B17WMDE_webpack_prototype&devMode"
```
gbirke: I don't understand why this is necessary and it makes the preview slightly inconsistent. Could you leave it as-is or would it break something?

shared/event_logging_tracker.js
```diff
@@ -46,6 +47,10 @@ export class EventLoggingTracker {
 		this.trackViewportData( this.bannerName, dimensionData, trackingRatio );
 	}
 
+	trackNamespaceEvent( dimensionData, trackingRatio = 1 ) {
```
gbirke: We want to track when a banner was not show because it was in the "wrong" namespace, or a "namespace where no banners are allowed". Could you rename this to something like `trackNamespacePreventedBannerDisplayEvent`, `trackNoBannerShownInNamespaceEvent`, or `trackBannerNotAllowedInNamespaceEvent`? I'm thinking of our future selves wondering what a "namespace event" is and having to jump from this implementation to the docblock in the interface ...

gbirke (changes requested):

campaign_info.toml
Abban: Oh, that was a left over from testing. Thanks for spotting it.

gbirke (approved):

moiikana:
rebased main

PR #1161, opened 2022-12-12
C22_WMDE_Test_21

moiikana:
ask for address type
https://phabricator.wikimedia.org/T324966


components/MultistepDonationForm/forms/AddressType.jsx
```diff
@@ -18,6 +18,9 @@ export default function AddressType( props ) {
 		props.trackBannerEvent( 'address-type-form-page-shown' );
 		if ( paymentMethod === PaymentMethods.DIRECT_DEBIT.value ) {
 			setDisabledAddressTypes( [ AddressTypes.NO.value ] );
+			if ( addressType === AddressTypes.NO.value ) {
```
Abban, line 21: I'm guessing this is to cover a use case where the user goes back and changes their payment method?

If so I think it'd be better in that the address type gets set to null than full because when there's 3 options we'd be picking one for the user

Abban (approved):

PR #1180, opened 2023-01-10
Re-enable dependabot 2023

Sperling-0:
- Re-enable dependabot
    
- Ticket: https://phabricator.wikimedia.org/T326550


Abban:
I think if we're going to rebuild in Vue this year that we should leave this one off?

